PR #20, opened 2022-07-20
Don't blow away package manifest when updating

### LINKED ISSUE #18: Don't blow away package manifest when updating version
Currently, when updating the version for a package manifest, we do not do a good job of preserving the original version of the manifest. First, `readManifestDependencyFields` includes dependency fields that aren't in the original manifest, and second, `readManifest` only includes manifest fields that we care about and discards the rest. Obviously, this is no bueno, and we should fix this.

mcmire:
When we read a package manifest, we need a representation of the
manifest that we can make reliable assumptions about. To do this, we
disregard fields that we don't care about and we fill in default values
for fields that we _do_ care about but haven't been explicitly provided.

The problem is that when updating the version of a package, we are using
a modified version of this representation of the manifest. That isn't
good, because it means that the package's manifest ends up getting
fundamentally changed.

This commit fixes this by storing the original representation of the
manifest when it is read. This representation is then used when the
version is updated instead of the "parsed" representation. This does
make the parsed representation immediately out of date, but that doesn't
really matter.

---

Fixes #18.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   },
   "dependencies": {
     "@metamask/action-utils": "^0.0.2",
-    "@metamask/utils": "^2.0.0",
+    "@metamask/utils": "^2.1.0",
     "debug": "^4.3.4",
     "execa": "^5.0.0",
     "glob": "^8.0.3",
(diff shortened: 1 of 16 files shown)


Gudahtt (approved):
LGTM!

PR #21, opened 2022-07-22
Simplify monorepo-workflow-operations

mcmire:
The tests for `monorepo-workflow-operations.ts` are not as readable or
maintainable as I'd like them to be. This commit attempts to fix that.

* While there isn't a whole lot we can do in terms of the scenarios
  we're testing because the function makes a lot of calls and the logic
  inside of it is somewhat complicated, we can at least move
  code that is responsible for parsing the TODAY environment variable,
  building a ReleasePlan object from the release spec, and applying the
  updates to the repos themselves out into separate files. This
  simplifies the test data and removes a few tests entirely.
* We also simplify the structure of the tests so that we aren't using so
  many nested `describe` blocks. This ends up being very difficult to
  keep straight, so the flattened layout here makes it a little more
  palatable.
* We also simplify the setup code for each test. Currently we are
  mocking all of the dependencies for `followMonorepoWorkflow` in one
  go, but we're doing so in a way that forces the reader to wade through
  a bunch of type definitions. That isn't really that helpful. The most
  complicated part of reading the tests for `followMonorepoWorkflow`
  isn't the dependencies — it's the logic. So we take all of the
  decision points we have to make in the implementation and represent
  those as options to our setup function in the tests so it's as clear
  as possible which exact scenario is being tested just by reading the
  test.
* Finally while we're moving things around, we can also move the check
  we perform to ensure that the new version of a package is greater than
  its existing version so that it is earlier in the entire workflow,
  i.e. from building the release plan to the validating the release
  specification. This also reduces the number of tests we have to write.


src/monorepo-workflow-operations.ts
```diff
@@ -137,118 +107,15 @@ export async function followMonorepoWorkflow({
...
-  ).map((packageName) => {
-    const pkg = project.workspacePackages[packageName];
-    const versionSpecifier = releaseSpecification.packages[packageName];
-    const currentVersion = pkg.validatedManifest.version;
-    let newVersion: SemVer;
-
-    if (versionSpecifier instanceof SemVer) {
-      const comparison = versionSpecifier.compare(currentVersion);
-
-      if (comparison === 0) {
```
mcmire, line 191: I've moved these checks to when the release spec is validated, as it makes more sense there.

src/initial-parameters.ts
```diff
@@ -32,6 +36,11 @@ export async function determineInitialParameters(
           project.rootPackage.validatedManifest.name.replace('/', '__'),
         )
       : path.resolve(cwd, inputs.tempDirectory);
+  const parsedTodayTimestamp =
```
mcmire, line 39: This is moved from `monorepo-workflow-operations`, where it was done on the fly.

src/release-plan.ts
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * @param args - The arguments.
+ * @param args.project - Information about the whole project (e.g., names of
+ * packages and where they can found).
+ * @param args.releaseSpecification - A parsed version of the release spec
+ * entered by the user.
+ * @param args.today - The current date.
+ * @returns A promise for information about the new release.
+ */
+export async function planRelease({
```
mcmire, line 56: This whole file was moved from `monorepo-workflow-operations`.

src/release-plan.ts
```diff
@@ -0,0 +1,129 @@
...
+  const workspaceReleasePlans: PackageReleasePlan[] = Object.keys(
+    releaseSpecification.packages,
+  ).map((packageName) => {
+    const pkg = project.workspacePackages[packageName];
+    const versionSpecifier = releaseSpecification.packages[packageName];
+    const currentVersion = pkg.validatedManifest.version;
+    const newVersion =
+      versionSpecifier instanceof SemVer
+        ? versionSpecifier
+        : new SemVer(currentVersion.toString()).inc(versionSpecifier);
```
mcmire, line 87: The checks that used to be here around the new version (formerly in `monorepo-workflow-operations.ts`) have been moved to `validateReleaseSpecification` in `release-specification.ts`.

src/repo.ts
```diff
@@ -78,3 +78,37 @@ export async function getRepositoryHttpsUrl(
...
+ * creates a new Git commit which carries the name of the new release.
+ * 2. Creates a new branch pointed to that commit (which also carries the name
+ * of the new release).
+ * 3. Switches to that branch.
+ *
+ * @param repositoryDirectoryPath - The path to the repository directory.
+ * @param releaseName - The name of the release, which will be used to name the
+ * commit and the branch.
+ */
+export async function captureChangesInReleaseBranch(
```
mcmire, line 95: This got moved from `workflow-operations.ts`, as it really pertains to a repo.

tests/unit/helpers.ts
```diff
@@ -104,58 +110,6 @@ export function buildMockPackage(
...
 
-/**
- * Builds a package for use in tests which is designed to be the root package of
- * a monorepo.
- *
- * @param args - The name of the package (optional), the version of the package
- * (optional) and the properties that will go into the object (optional).
- * @returns The mock Package object.
- */
-export function buildMockMonorepoRootPackage(
```
mcmire, line 115: This is now no longer used entirely.

tests/setupAfterEnv.ts
```diff
@@ -77,4 +80,49 @@ expect.extend({
...
+   * Tests that the given promise is rejected within a certain amount of time
+   * (which defaults to the time that Jest tests wait before timing out as
+   * configured in the Jest configuration file).
+   *
+   * Inspired by <https://stackoverflow.com/a/68409467/260771>.
+   *
+   * @param promise - The promise to test.
+   * @returns The result of the matcher.
+   */
+  async toBeRejected(promise: Promise<any>) {
```
Gudahtt, line 94: Hmm, this is neat, but what advantages does this have over using `.rejects` and relying upon the built-in timeout?

I guess this would be useful in cases where the timeout is disabled or is very long (e.g. the extension e2e tests), but we're not in that situation here. Or when you want quite a short timeout just for this one step.

tests/setupAfterEnv.ts
```diff
@@ -47,7 +48,9 @@ expect.extend({
    */
   async toResolve(promise: Promise<any>) {
     if (this.isNot) {
-      throw new Error('Using `.not.toResolve(...)` is not supported.');
+      throw new Error(
+        'Using `.not.toResolve()` is not supported. Use .toReject() instead.',
```
Gudahtt, line 52: Was this meant to be `toBeRejected`?

`toReject` is more in-line with `toResolve` though.

tests/setupAfterEnv.ts
mcmire: Yeah, I guess instead of using this matcher I could use `await expect(...).rejects.toThrow(expect.anything())`. It's symmetrical with `toResolve`, which I like, but I guess it's not completely necessary. Okay, I can make a replacement.

Gudahtt: I'm ok with keeping it as well if you want! It has some utility even if we don't strictly need it here. I was just making sure I wasn't missing something

mcmire: It's okay. We can always add it back!

src/monorepo-workflow-operations.ts
```diff
@@ -75,29 +43,31 @@ function getToday() {
...
   tempDirectoryPath: string;
   firstRemovingExistingReleaseSpecification: boolean;
+  today: Date;
   stdout: Pick<WriteStream, 'write'>;
   stderr: Pick<WriteStream, 'write'>;
 }) {
   const releaseSpecificationPath = path.join(tempDirectoryPath, 'RELEASE_SPEC');
 
-  if (firstRemovingExistingReleaseSpecification) {
-    await promisifiedRimraf(releaseSpecificationPath);
```
Gudahtt, line 97: I guess it's fine to leave it around and let it get overwritten. That makes sense. I assume that was the intention here.

mcmire: Right, exactly!

src/release-plan.ts
Gudahtt: Good idea, that makes much more sense

tests/setupAfterEnv.ts
mcmire: I updated this error message since I removed `toBeRejected`.

src/monorepo-workflow-operations.test.ts
```diff
@@ -1,1788 +1,1214 @@
...
+            project,
+            tempDirectoryPath: sandbox.directoryPath,
+            firstRemovingExistingReleaseSpecification: false,
+            today,
+            stdout,
+            stderr,
           });
+
+          expect(stdout.data()[0]).toMatch(
+            /^A template has been generated that specifies this release/u,
```
Gudahtt, line 489: It looks like we lost the test for ensuring this file gets created. Unless I'm missing it

src/monorepo-workflow-operations.test.ts
```diff
@@ -1,1788 +1,1214 @@
...
-                    path.join(sandbox.directoryPath, 'RELEASE_SPEC'),
-                    'utf8',
-                  ),
-                ).rejects.toThrow(/^ENOENT: no such file or directory/u);
-              });
+          expect(await fileExists(releaseSpecificationPath)).toBe(false);
+        });
+      });
+
+      it('does not remove the release spec file', async () => {
```
Gudahtt, line 602: Hmm, this looks out of place. This describe block is for when the release spec file exists, but it doesn't exist here

src/monorepo-workflow-operations.test.ts
```diff
@@ -1,1788 +1,1214 @@
...
-                  packageReleasePlan: {
-                    package: project.workspacePackages.d,
-                    newVersion: '1.2.3',
-                    shouldUpdateChangelog: true,
-                  },
-                  stderr,
-                });
-              });
-            });
+    describe('when firstRemovingExistingReleaseSpecification is false and the release spec file already exists', () => {
```
Gudahtt, line 518: Thoughts on whether we should have a test to ensure it doesn't open the editor?

src/monorepo-workflow-operations.test.ts
```diff
@@ -1,1788 +1,1214 @@
...
+            stdout,
+            stderr,
+          });
+
+          expect(await fileExists(releaseSpecificationPath)).toBe(true);
+        });
+      });
+    });
+
+    describe('when firstRemovingExistingReleaseSpecification is true, the release spec file already exists, an editor is available, and editing will succeed', () => {
```
Gudahtt, line 920: Similarly, is there a way for us to ensure it's following the new release spec here rather than the old one?

Gudahtt (commented):
These improvements look good! I see why you removed the nested describe blocks.

I had a few questions about the tests but aside from that I've finished reviewing it all.

src/monorepo-workflow-operations.test.ts
mcmire: Hmm, yes, it looks like tests are missing for what happens if `validateReleaseSpecification`, `planRelease` or `executeReleasePlan` throws (which would cause the file to not get removed, in the case in which it is created). I'll add those.

src/monorepo-workflow-operations.test.ts
mcmire: Oops! Looks like copypasta. Removed in [9c04d44](https://github.com/MetaMask/create-release-branch/pull/21/commits/9c04d444e2042c16970d3d4dea601af2968b72e7).

src/monorepo-workflow-operations.test.ts
mcmire: Okay, added in [aca7cce](https://github.com/MetaMask/create-release-branch/pull/21/commits/aca7cce4a026221f9da66c99bf1cfa66710579c5).

src/monorepo-workflow-operations.test.ts
mcmire: Ah, makes sense. Added in [fbc5bf4](https://github.com/MetaMask/create-release-branch/pull/21/commits/fbc5bf415790580d82b360e9971d38a78b63c448).

src/monorepo-workflow-operations.test.ts
mcmire: Yup, makes sense too. Added in [fbc5bf4](https://github.com/MetaMask/create-release-branch/pull/21/commits/fbc5bf415790580d82b360e9971d38a78b63c448).

Gudahtt (approved):
LGTM!

PR #22, opened 2022-07-22
Reformat root version to include the release no.

### LINKED ISSUE #16: Monorepo independent flow: Include build number in release version
It is possible to create multiple release branches on the same day. To do this, the release version must be unique. We can do this by including a build number in the release version. Right now the release version is formatted as `YYYY.MM.DD`. We want to change this to `YYYYMMDD.BUILD_NUMBER.0`, and we want to increment the `BUILD_NUMBER` for every fresh run of this tool.

mcmire:
This commit changes how the root version is bumped per release so that
instead of looking like `<yyyy>.<mm>.<dd>`, it looks like
`<yyyy><mm><dd>.<release-number>.0`. That is, the root version
consists of the date that the release was created + a release number,
which is incremented every time a new release is issued.

---

Fixes #16.

### DIFF
--- a/src/initial-parameters.test.ts
+++ b/src/initial-parameters.test.ts
@@ -46,7 +46,7 @@ describe('initial-parameters', () => {
         project,
         tempDirectoryPath: '/path/to/temp',
         reset: true,
-        today: new Date('2022-06-22'),
+        today: new Date(2022, 5, 22),
       });
     });
 
@@ -125,9 +125,81 @@ describe('initial-parameters', () => {
       );
     });
 
-    it('uses the current date if the TODAY environment variable was not provided', async () => {
+    it('returns initial parameters including reset: false, derived from a command-line argument of "--reset true"', async () => {
       const project = buildMockProject();
-      const today = new Date('2022-01-01');
+      when(jest.spyOn(commandLineArgumentsModule, 'readCommandLineArguments'))
+        .calledWith(['arg1', 'arg2'])
+        .mockResolvedValue({
+          projectDirectory: '/path/to/project',
+          tempDirectory: '/path/to/temp',
+          reset: true,
+        });
+      jest
+        .spyOn(envModule, 'getEnvironmentVariables')
+        .mockReturnValue({ TODAY: undefined, EDITOR: undefined });
+      when(jest.spyOn(projectModule, 'readProject'))
+        .calledWith('/path/to/project')
+        .mockResolvedValue(project);
+
+      const config = await determineInitialParameters(
+        ['arg1', 'arg2'],
+        '/path/to/somewhere',
+      );
+
+      expect(config.reset).toBe(true);
+    });
+
+    it('returns initial parameters including reset: false, derived from a command-line argument of "--reset false"', async () => {
+      const project = buildMockProject();
+      when(jest.spyOn(commandLineArgumentsModule, 'readCommandLineArguments'))
+        .calledWith(['arg1', 'arg2'])
+        .mockResolvedValue({
+          projectDirectory: '/path/to/project',
+          tempDirectory: '/path/to/temp',
+          reset: false,
+        });
+      jest
+        .spyOn(envModule, 'getEnvironmentVariables')
+        .mockReturnValue({ TODAY: undefined, EDITOR: undefined });
+      when(jest.spyOn(projectModule, 'readProject'))
+        .calledWith('/path/to/project')
+        .mockResolvedValue(project);
+
+      const config = await determineInitialParameters(
+        ['arg1', 'arg2'],
+        '/path/to/somewhere',
+      );
+
+      expect(config.reset).toBe(false);
+    });
+
+    it("returns initial parameters including today's date, derived from the TODAY environment variable", async () => {
+      const project = buildMockProject();
+      when(jest.spyOn(commandLineArgumentsModule, 'readCommandLineArguments'))
+        .calledWith(['arg1', 'arg2'])
+        .mockResolvedValue({
+          projectDirectory: '/path/to/project',
+          tempDirectory: '/path/to/temp',
+          reset: true,
+        });
+      jest
+        .spyOn(envModule, 'getEnvironmentVariables')
+        .mockReturnValue({ TODAY: '2022-01-01', EDITOR: undefined });
+      when(jest.spyOn(projectModule, 'readProject'))
+        .calledWith('/path/to/project')
+        .mockResolvedValue(project);
+
+      const config = await determineInitialParameters(
+        ['arg1', 'arg2'],
+        '/path/to/somewhere',
+      );
+
+      expect(config.today).toStrictEqual(new Date(2022, 0, 1));
+    });
+
+    it('uses the current date if TODAY is undefined', async () => {
+      const project = buildMockProject();
+      const today = new Date(2022, 0, 1);
       when(jest.spyOn(commandLineArgumentsModule, 'readCommandLineArguments'))
         .calledWith(['arg1', 'arg2'])
         .mockResolvedValue({
@@ -153,7 +225,7 @@ describe('initial-parameters', () => {
 
     it('uses the current date if TODAY is not a parsable date', async () => {
       const project = buildMockProject();
-      const today = new Date('2022-01-01');
+      const today = new Date(2022, 0, 1);
       when(jest.spyOn(commandLineArgumentsModule, 'readCommandLineArguments'))
         .calledWith(['arg1', 'arg2'])
         .mockResolvedValue({
--- a/src/initial-parameters.test.ts
+++ b/src/initial-parameters.test.ts
@@ -124,7 +124,7 @@ describe('initial-parameters', () => {
       );
     });
 
-    it('returns initial parameters including reset: false, derived from a command-line argument of "--reset true"', async () => {
+    it('returns initial parameters including reset: true, derived from a command-line argument of "--reset true"', async () => {
       const project = buildMockProject();
       when(jest.spyOn(commandLineArgumentsModule, 'readCommandLineArguments'))
         .calledWith(['arg1', 'arg2'])
--- a/src/initial-parameters.test.ts
+++ b/src/initial-parameters.test.ts
@@ -32,7 +32,7 @@ describe('initial-parameters', () => {
         });
       jest
         .spyOn(envModule, 'getEnvironmentVariables')
-        .mockReturnValue({ TODAY: '2022-06-22', EDITOR: undefined });
+        .mockReturnValue({ EDITOR: undefined });
       when(jest.spyOn(projectModule, 'readProject'))
         .calledWith('/path/to/project')
         .mockResolvedValue(project);
@@ -46,7 +46,6 @@ describe('initial-parameters', () => {
         project,
         tempDirectoryPath: '/path/to/temp',
         reset: true,
-        today: new Date(2022, 5, 22),
       });
     });
 
@@ -63,7 +62,7 @@ describe('initial-parameters', () => {
         });
       jest
         .spyOn(envModule, 'getEnvironmentVariables')
-        .mockReturnValue({ TODAY: undefined, EDITOR: undefined });
+        .mockReturnValue({ EDITOR: undefined });
       const readProjectSpy = jest
         .spyOn(projectModule, 'readProject')
         .mockResolvedValue(project);
@@ -84,7 +83,7 @@ describe('initial-parameters', () => {
         });
       jest
         .spyOn(envModule, 'getEnvironmentVariables')
-        .mockReturnValue({ TODAY: undefined, EDITOR: undefined });
+        .mockReturnValue({ EDITOR: undefined });
       when(jest.spyOn(projectModule, 'readProject'))
         .calledWith('/path/to/project')
         .mockResolvedValue(project);
@@ -110,7 +109,7 @@ describe('initial-parameters', () => {
         });
       jest
         .spyOn(envModule, 'getEnvironmentVariables')
-        .mockReturnValue({ TODAY: undefined, EDITOR: undefined });
+        .mockReturnValue({ EDITOR: undefined });
       when(jest.spyOn(projectModule, 'readProject'))
         .calledWith('/path/to/project')
         .mockResolvedValue(project);
@@ -136,7 +135,7 @@ describe('initial-parameters', () => {
         });
       jest
         .spyOn(envModule, 'getEnvironmentVariables')
-        .mockReturnValue({ TODAY: undefined, EDITOR: undefined });
+        .mockReturnValue({ EDITOR: undefined });
       when(jest.spyOn(projectModule, 'readProject'))
         .calledWith('/path/to/project')
         .mockResolvedValue(project);
@@ -160,7 +159,7 @@ describe('initial-parameters', () => {
         });
       jest
         .spyOn(envModule, 'getEnvironmentVariables')
-        .mockReturnValue({ TODAY: undefined, EDITOR: undefined });
+        .mockReturnValue({ EDITOR: undefined });
       when(jest.spyOn(projectModule, 'readProject'))
         .calledWith('/path/to/project')
         .mockResolvedValue(project);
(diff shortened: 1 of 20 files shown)


src/initial-parameters.ts
```diff
@@ -37,7 +38,7 @@ export async function determineInitialParameters(
         )
       : path.resolve(cwd, inputs.tempDirectory);
   const parsedTodayTimestamp =
-    TODAY === undefined ? NaN : new Date(TODAY).getTime();
+    TODAY === undefined ? NaN : parseDateAsISO(TODAY).getTime();
```
mcmire, line 41: Here `TODAY` is an ISO string. We could pass it straight to the Date constructor, but that would result in a Date in UTC rather than local time. We can work with this Date object using `.getUTCFullYear()`, `.getUTCMonth()`, and `.getUTCDate()`, but I'd rather not place that burden on those of us who will maintain this. To fix this I'm using `parseISO` from `date-fns`, which will give us back a time in local time (e.g. `parseDateAsISO('2022-08-25')` => `new Date(2022, 7, 25)`). I guess ultimately it doesn't matter since we use `formatDate` (also from `date-fns`) later on to turn this back into some kind of date string, but when debugging this it seems more straightforward to see a local time here.

src/project.test.ts
```diff
@@ -66,31 +66,68 @@ describe('project', () => {
...
           isMonorepo: true,
+          releaseInfo: {
+            releaseDate: new Date(2022, 6, 22),
+            releaseNumber: 1234,
+          },
         });
       });
     });
 
-    it('collects information about a polyrepo project', async () => {
```
mcmire, line 73: I decided to drop the polyrepo test from this because `readProject` doesn't technically support polyrepo projects right now and that is more true with this PR than before. I think this is something we can fix in a later PR.

mcmire:
This is ready for review.

src/release-plan.ts
```diff
@@ -62,12 +62,11 @@ export async function planRelease({
...
 }): Promise<ReleasePlan> {
-  const newReleaseName = today.toISOString().replace(/T.+$/u, '');
-  const newRootVersion = [
-    today.getUTCFullYear(),
-    today.getUTCMonth() + 1,
-    today.getUTCDate(),
-  ].join('.');
+  const newReleaseDate = formatDateAsISO(today, {
+    representation: 'date',
+  }).replace(/\D+/gu, '');
```
Gudahtt, line 67: Is this step meant to get rid of the dashes?

Gudahtt: I've just discovered that this can be done by setting the `format` option to `basic`

src/monorepo-workflow-operations.test.ts
```diff
@@ -89,21 +89,19 @@ function buildMockReleaseSpecification({
...
+ * @param overrides.releaseNumber - The number of the release.
  * @param overrides.packages - Information about all of the packages in the
  * project. For a polyrepo, this consists of the self-same package; for a
  * monorepo it consists of the root package and any workspace packages.
  * @returns The mock release specification.
  */
 function buildMockReleasePlan({
-  releaseName = 'release-name',
+  releaseDate = new Date(),
+  releaseNumber = 1,
```
Gudahtt, line 101: Curious why this number starts at one rather than zero.

src/release-plan.ts
mcmire: Yeah, it's meant to get rid of the dashes. Thanks for the tip, I didn't look hard enough at the docs I guess 🤦🏻 Fixed in [a4f69eb](https://github.com/MetaMask/create-release-branch/pull/22/commits/a4f69eb6da28e0e34ae24d6156482e19cd232c37).

src/project.ts
```diff
@@ -21,13 +21,84 @@ export interface Project {
...
   isMonorepo: boolean;
+  releaseInfo: ReleaseInfo;
+}
+
+/**
+ * Information about the release of the root package of a monorepo extracted
+ * from its version string.
+ *
+ * @property releaseDate - The release date.
+ * @property releaseNumber - The release number (starting from 1).
```
Gudahtt, line 32: Any reason to start at one rather than zero? 

I guess it is conventional to start at zero with the "RC__" syntax. But I've more typically seen it start at zero when the build number is just a part of the semver version. But we have both representations in different places.

src/release-plan.ts
```diff
@@ -62,12 +62,12 @@ export async function planRelease({
...
-  const newRootVersion = [
-    today.getUTCFullYear(),
-    today.getUTCMonth() + 1,
-    today.getUTCDate(),
-  ].join('.');
+  const newReleaseDate = formatDateAsISO(today, {
+    format: 'basic',
+    representation: 'date',
+  });
+  const newReleaseNumber = project.releaseInfo.releaseNumber + 1;
```
Gudahtt, line 69: Hmm, shouldn't the release number only be increased if the date stays the same?

Gudahtt: Currently it seems like it's impossible to set the release number to 1 for a new release. It always increases the previous release number, and the previous release number can't be 1, so the first release will have a release number of 2 and it'll go up from there.

mcmire: > Hmm, shouldn't the release number only be increased if the date stays the same?

It could. My thought is that from a programmatic perspective, instead of using the date to uniquely identify a release, we use a number that is always incremented by 1 for each new release to do so. Colloquially, we'd switch it around and generally use the date as a more human-readable identifier, but then use the release number to disambiguate multiple releases on the same day. From an implementation perspective, it simplifies things a bit because it means release branches would always be formatted the same and so would release notes.

For instance, say we have two releases that end up going out today. So the root package version would be `20220826.100.0`. When the branch is merged, the resulting GitHub release would have the following release notes:

> # August 26, 2022 (R100)
>
> ## @metamask/a 1.2.0
>
> ### Added
>
> * Add feature A (#94)
>
> ## @metamask/b 0.7.0
>
> ### Added
>
> * Add feature B (#91)
>
> ### Changed
>
> * Bump `@metamask/a` to 1.2.0 (#90)

Then we'd publish version `20220826.101.0`, and the GitHub release would have the following release notes:

> # August 26, 2022 (R101)
>
> ## @metamask/a 1.2.1
>
> ### Fixed
>
> * Fix some bug (#104)
>
> ## @metamask/b 0.7.1
>
> ### Changed
>
> * Bump `@metamask/a` to 1.2.1 (#102)

If we only bumped the middle number when the date changed, that's fine, but we'd have to work out how to refer to a release which is the only instance on a particular day and one which is not.

Going back to the example, say again we have two releases that end up going out today. We release `20220826.1.0`, but what would the GitHub release notes look like? Maybe we use the same number:

> # August 26, 2022 (1)
>
> ... release notes go here ...

and then the release notes for the second release would be:

> # August 26, 2022 (2)
>
> ... release notes go here ...

So, okay, that works fine. But multi-release dates are the exception, not the rule. So in reality, it means we'll have a lot of release notes that look like:

> # August 26, 2022 (1)
>
> ...
>
> # September 1, 2022 (1)
>
> ...
>
> # October 17, 2022 (1)
>
> ...

If we use an ever-increasing release number, we wouldn't have to add extra logic to increment the number only when the date changes, and we could generate the release notes in a way that's consistent and wouldn't look weird:

> # August 26, 2022 (R100)
>
> ...
>
> # September 1, 2022 (R101)
>
> ...
>
> # October 17, 2022 (R102)
>
> ...
>
> # October 17, 2022 (R103)
>
> ...

src/project.ts
mcmire: My thought is that we could use the release number to generate the release notes, so we say e.g. "August 26, 2022 (R1)" for the very first release instead of "August 26, 2022 (R0)". For that reason 1 seems like a more natural starting number than 0, but we can technically do whatever we want here.

src/release-plan.ts
mcmire: > My thought is that from a programmatic perspective, instead of using the date to uniquely identify a release, we use a number that is always incremented by 1 for each new release to do so.

Quoting myself here, but I realize that I could switch things around so instead of `<date>.<release-number>.0` it could be `<release-number>.<date>.0`. That might seem a little more SemVer-y because it means that the first number would act more like the "major" number instead of the second.

Gudahtt: Oh OK, I thought it's only purpose was to allow releasing twice on a day if we need to.

If we're going with an ever-increasing number, which is fine, do we need the date? We could just use the release number.

Gudahtt: >I thought it's only purpose was to allow releasing twice on a day if we need to.
Oh, and backporting. It also allowed for that.

That's also compatible with dropping the date and just using the incrementing release number, except that we'd need to use an additional digit in the version for that, like the patch version.

mcmire: > Currently it seems like it's impossible to set the release number to 1 for a new release. It always increases the previous release number, and the previous release number can't be 1, so the first release will have a release number of 2 and it'll go up from there.

Ah. Good point. Maybe to fix this the tool could look at the Git tags and if there aren't any that it recognizes as release tags, it assumes that this is the very first release. In that case it check to make sure that the release number is 1 (or whatever we want the starting value to be) and leave it alone, otherwise it increments it. I also realize I have functional tests for the "very first release" case, but they are wrong. I'll have to fix this in another PR since #23 will introduce the use of Git tags to filter the set of packages that show up in the release spec.

Gudahtt: To come back to the overall goals a bit, here is what I think we want from the version:

* A number that goes up.
* No constraints on how often we can release.
* A way to release a new version that is in-between two existing release versions (i.e. a backport release), in case we need to update an older package because it's used by a project that can't take on the work of upgrading to latest yet.
* SemVer compatibility on the _syntax_ level, so that tools that assume SemVer will work somewhat-correctly. Semantically we won't be following SemVer rules, but diverging on syntax seems like inviting an unnecessary headache.

Not that we need to solve all of these problems in this PR, I just mean that's what the versioning scheme needs to provide us.

And lastly, as a nice-to-have, it'd be better if this number was easy for us to understand and talk about/read. Simple is better in case we need to say it out-loud. That's all that I see in terms of requirements.

* `<date>.<release number>.0` where the release number resets each day fits all of these goals.
  * In printing the name, and in talking about it generally, we can omit the release number unless we have multiple releases that day. So in your example, we can drop the `(1)`'s in the changelog.
  * We can increment the release number for multiple releases in a day, OR for backports. It solves both of those problems
    * I'm just now realizing that this doesn't allow for a way of backporting a release on the same day it went out (e.g. R1 is published, then R2, then I want to backport to R1). Hmm. Seems unlikely but not ideal. We could use the patch number for backports here to eliminate this edge case.
* `<date>.<release number>.0` where the release number constantly goes up also solves all of these problems, as long as we can use the patch number for backports.
* `<release number>.<backport number>.0` also solves this problem

I'm OK with any of these solutions, and they're all readable-enough for me. But the ever-increasing release number plus a date option does seem like the worse of these options. We don't need two numbers that go up. I think in practice we'd just end up referring to it by the date or the release number.

mcmire: Okay, I'm hearing two concerns: the backporting workflow and the utility of the number (or the date) in the version string.

To me, there are two goals of the root version: to provide a unique name for a release (as version strings always do) and to encode information that can be used by the `publish-release` action. Right now, when the `publish-release` action constructs a title for a new GitHub release, it doesn't need to read any action parameters or environment variables to do so, it can merely reuse the version string of the root package. My idea for the core monorepo (so far) has been to go one step further and extract the date and number from the version string, then format these two things into a more "pretty" title (e.g. translate `20220906.123.0` into "Sept. 6, 2022 (R123)").

Hopefully this provides a little context for the answers below.

---

As for the release number:

> the ever-increasing release number plus a date option does seem like the worse of these options. We don't need two numbers that go up. I think in practice we'd just end up referring to it by the date or the release number. Which... is fine, but, I'm curious what advantages you see in it over the others.

Sure, good question. I think the advantages came down to a slightly simpler implementation. It wouldn't be difficult to add logic that would decide when to bump that number, but I think we'd have to give a different name to the concept so that we can document it well, as "release number" might be a bit confusing. These are minor problems, but then also:

> In printing the name, and in talking about it generally, we can omit the release number unless we have multiple releases that day. So in your example, we can drop the `(1)`'s in the changelog.

If the "release number" resets each day, then the title for the first release for a given day would not be suffixed with "(1)", and subsequent releases for the same day _would_ be suffixed — presumably, "(2)", "(3)", etc. I believe I wanted to avoid this, so that the title would be formatted the same way regardless of whether it was the first release in a sequence or the only release in a sequence, and therefore it would be easier to talk about.

---

As for backporting:

> [...] here is what I think we want from the version [...]
>
> - A way to release a new version that is in-between two existing release versions (i.e. a backport release), in case we need to update an older package because it's used by a project that can't take on the work of upgrading to latest yet.

My thought is that the date would always represent the release date, not the date that a branch was cut. So while a backport release would functionally sit in between two other versions, the name for that release wouldn't. For example, if we had two versions `20220501.1.0` and `20220801.1.0`, and we wanted to backport a fix on top of `20220501.1.0`, the resulting version string would not be `20220501.1.1` but would still be `<current date>.<number>.0`. Again, this is so that the title of the resulting GitHub release would include the current date and not an older date.

That said, I hadn't thought about this case too much, and I can see how useful it would be to know whether a particular version represents a backport of an existing version.

---

Putting these two concerns together, it appears that the final option you proposed, `<release number>.<backport number>.0`, would probably be the best one. Would the "release number" increase with each release or not? Based on your comments I'm inclined to think it would not, so assuming this is the case:

* From a documentation/implementation perspective, I still think the first number needs a different name. Perhaps we make it simple and call the whole format `<major>.<minor>.<patch>` (where `patch` is always 0), or even `<major>.<patch>.0` (where `major` is understood to mean "a number that usually increases but may not" and `patch` is understood to mean "a number that increases with a backport release, like one might do for a hotfix release elsewhere").
* If the first number usually increases, except for backport releases, does the tool need to take a `--backport` option? Without this option, the current version would be incremented by bumping the first number, but with this option, it would be incremented by bumping only the second number.
* If we removed the date from the version string, then it wouldn't make sense to translate that version string into something human-readable when constructing the GitHub release title. Perhaps we go back to using the version string as the title as we do now?

Gudahtt: >it appears that the final option you proposed, <release number>.<backport number>.0, would probably be the best one. Would the "release number" increase with each release or not?

Ah, sorry, with that suggestion I meant to suggest an ever-increasing release number as you've used here. Forgot to specify that. that was meant to represent [this comment](https://github.com/MetaMask/create-release-branch/pull/22#discussion_r956488979).

Gudahtt: >If the first number usually increases, except for backport releases, does the tool need to take a --backport option? Without this option, the current version would be incremented by bumping the first number, but with this option, it would be incremented by bumping only the second number.

Yes, I think so. I can make an issue for this

>If we removed the date from the version string, then it wouldn't make sense to translate that version string into something human-readable when constructing the GitHub release title. Perhaps we go back to using the version string as the title as we do now?

Makes sense!

Gudahtt: I created a ticket for the backport support here: https://github.com/MetaMask/create-release-branch/issues/32

mcmire: Following our discussion in standup, for `<release number>.<backport number>.0` we would want only `release number` to increase for normal releases and only `backport number` to increase for backport releases.

mcmire:
I wasn't able to get to this today. I may have some time this weekend to make these changes but can't promise anything. Sorry :(

src/initial-parameters.ts
```diff
@@ -1,14 +1,12 @@
...
 import path from 'path';
-import { getEnvironmentVariables } from './env';
 import { readCommandLineArguments } from './command-line-arguments';
 import { readProject, Project } from './project';
 
 interface InitialParameters {
   project: Project;
   tempDirectoryPath: string;
   reset: boolean;
-  today: Date;
```
mcmire, line 11: There's no need to read the date from `TODAY` nor pass this down to anywhere, as a date isn't used anymore to calculate the version string. So it gets removed from everywhere.

src/project.ts
```diff
@@ -3,6 +3,25 @@ import glob from 'glob';
 import { Package, readPackage } from './package';
 import { PackageManifestFieldNames } from './package-manifest';
 import { getRepositoryHttpsUrl } from './repo';
+import { SemVer } from './semver';
+
+/**
+ * The release version of the root package of a monorepo extracted from its
+ * version string.
+ *
+ * @property ordinaryNumber - The number assigned to the release if it
```
mcmire, line 12: I tossed around a few names in my head and went with "ordinary" as an opposite of "backport". Is this a good name?

src/release-plan.ts
mcmire: I've updated the version in a4f69eb...6f1274c.

src/project.ts
mcmire: This is now outdated following discussion+changes, resolving.

src/initial-parameters.ts
mcmire: Outdated due to changes, resolving.

mcmire:
This is now updated as per discussion around the release version.

src/project.ts
Gudahtt: Seems like a good name to me!

src/initial-parameters.test.ts
```diff
@@ -125,56 +124,52 @@ describe('initial-parameters', () => {
       );
     });
 
-    it('uses the current date if the TODAY environment variable was not provided', async () => {
+    it('returns initial parameters including reset: false, derived from a command-line argument of "--reset true"', async () => {
```
Gudahtt, line 127: Typo?
```suggestion
    it('returns initial parameters including reset: true, derived from a command-line argument of "--reset true"', async () => {
```

mcmire: Yup! Good catch

Gudahtt (approved):
LGTM!

PR #23, opened 2022-07-27
Update release spec to only list changed packages

### LINKED ISSUE #8: Monorepo flow: When generating a release spec, only list changed packages
A release spec, when generated, should only contain packages that have been changed since the last release, not all packages as it does now. This is because it is useless to publish a package that has no changes in it.

mcmire:
This PR adds a new property to the Package object,
`hasChangesSinceLatestRelease`. This property is computed when a package
is read from a project by running a diff between the last-created Git
tag associated with that package and whatever the HEAD commit happens to
be, then checking to see whether any of the files changed belong to that
package. If they do, then `hasChangesSinceLatestRelease` is true,
otherwise it's false. This property is then used to filter the list of
packages that are placed within the release spec template when it is
generated.

There are a couple of things to consider here:

* "the last-created Git tag associated with that package" — how do we
know this? How do we map a package's version to a tag? We have to
account for tags that were created by `action-create-release-pr` in the
past as well as the tags that this tool will create in the future. We
also have to know what kind of package this is — whether it's the root
package of a monorepo or a workspace package — because the set of
possible tags will differ. These differences are documented in
`readMonorepoRootPackage` and `readMonorepoWorkspacePackage`.
* What happens if a repo has no tags? Then all of the packages in that
repo are considered to have changed, as they have yet to receive their
initial release, so they will all be included in the release spec
template.

---

Fixes #8.

### DIFF
--- a/src/package.test.ts
+++ b/src/package.test.ts
@@ -67,6 +67,33 @@ describe('package', () => {
       });
     });
 
+    it('throws if a tag matching the current version does not exist', async () => {
+      jest
+        .spyOn(packageManifestModule, 'readPackageManifest')
+        .mockResolvedValue({
+          unvalidated: {},
+          validated: buildMockManifest({
+            name: 'some-package',
+            version: new SemVer('1.0.0'),
+          }),
+        });
+      when(jest.spyOn(repoModule, 'hasChangesInDirectorySinceGitTag'))
+        .calledWith('/path/to/project', '/path/to/package', 'v1.0.0')
+        .mockResolvedValue(true);
+
+      const promiseForPkg = readMonorepoRootPackage({
+        packageDirectoryPath: '/path/to/package',
+        projectDirectoryPath: '/path/to/project',
+        projectTagNames: ['some-tag'],
+      });
+
+      await expect(promiseForPkg).rejects.toThrow(
+        new Error(
+          'The package some-package has no Git tag for its current version 1.0.0 (expected "v1.0.0"), so this tool is unable to determine whether it should be included in this release. You will need to create a tag for this package in order to proceed.',
+        ),
+      );
+    });
+
     it("returns the fact that the package has been changed since its latest release, if a tag matching the current version exists and changes have been made to the package's directory since the tag", async () => {
       jest
         .spyOn(packageManifestModule, 'readPackageManifest')
@@ -198,6 +225,7 @@ describe('package', () => {
         packageDirectoryPath: '/path/to/package',
         projectDirectoryPath: '/path/to/project',
         projectTagNames: [],
+        rootPackageName: 'root-package',
         rootPackageVersion: new SemVer('5.0.0'),
       });
 
@@ -222,6 +250,7 @@ describe('package', () => {
         packageDirectoryPath: '/path/to/package',
         projectDirectoryPath: '/path/to/project',
         projectTagNames: [],
+        rootPackageName: 'root-package',
         rootPackageVersion: new SemVer('5.0.0'),
       });
 
@@ -229,6 +258,34 @@ describe('package', () => {
         unvalidatedManifest,
         validatedManifest,
       });
+    });
+
+    it('throws if a tag matching the current version does not exist', async () => {
+      const unvalidatedManifest = {};
+      const validatedManifest = buildMockManifest({
+        name: 'workspace-package',
+        version: new SemVer('1.0.0'),
+      });
+      when(jest.spyOn(packageManifestModule, 'readPackageManifest'))
+        .calledWith('/path/to/package/package.json')
+        .mockResolvedValue({
+          unvalidated: unvalidatedManifest,
+          validated: validatedManifest,
+        });
+
+      const promiseForPkg = readMonorepoWorkspacePackage({
+        packageDirectoryPath: '/path/to/package',
+        projectDirectoryPath: '/path/to/project',
+        projectTagNames: ['some-tag'],
+        rootPackageName: 'root-package',
+        rootPackageVersion: new SemVer('5.0.0'),
+      });
+
+      await expect(promiseForPkg).rejects.toThrow(
+        new Error(
+          'The workspace package workspace-package has no Git tag for its current version 1.0.0 (expected "workspace-package@1.0.0"), and the root package root-package has no Git tag for its current version 5.0.0 (expected "v5.0.0"), so this tool is unable to determine whether the workspace package should be included in this release. You will need to create tags for both of these packages in order to proceed.',
+        ),
+      );
     });
 
     it("returns the fact that the package has been changed since its latest release, if a tag matching the package name + version exists and changes have been made to the package's directory since the tag", async () => {
@@ -253,6 +310,7 @@ describe('package', () => {
         packageDirectoryPath: '/path/to/package',
         projectDirectoryPath: '/path/to/project',
         projectTagNames: ['@scope/some-package@1.0.0'],
+        rootPackageName: 'root-package',
         rootPackageVersion: new SemVer('5.0.0'),
       });
 
@@ -283,6 +341,7 @@ describe('package', () => {
         packageDirectoryPath: '/path/to/package',
         projectDirectoryPath: '/path/to/project',
         projectTagNames: ['@scope/some-package@1.0.0'],
+        rootPackageName: 'root-package',
         rootPackageVersion: new SemVer('5.0.0'),
       });
 
@@ -308,6 +367,7 @@ describe('package', () => {
         packageDirectoryPath: '/path/to/package',
         projectDirectoryPath: '/path/to/project',
         projectTagNames: ['v5.0.0'],
+        rootPackageName: 'root-package',
         rootPackageVersion: new SemVer('5.0.0'),
       });
 
@@ -333,6 +393,7 @@ describe('package', () => {
         packageDirectoryPath: '/path/to/package',
         projectDirectoryPath: '/path/to/project',
         projectTagNames: ['v5.0.0'],
+        rootPackageName: 'root-package',
         rootPackageVersion: new SemVer('5.0.0'),
       });
 
@@ -355,6 +416,7 @@ describe('package', () => {
         packageDirectoryPath: '/path/to/package',
         projectDirectoryPath: '/path/to/project',
         projectTagNames: [],
+        rootPackageName: 'root-package',
         rootPackageVersion: new SemVer('5.0.0'),
       });
 
--- a/src/package.test.ts
+++ b/src/package.test.ts
@@ -94,7 +94,7 @@ describe('package', () => {
       );
     });
 
-    it("returns the fact that the package has been changed since its latest release, if a tag matching the current version exists and changes have been made to the package's directory since the tag", async () => {
+    it("flags the package as having been changed since its latest release if a tag matching the current version exists and changes have been made to the package's directory since the tag", async () => {
       jest
         .spyOn(packageManifestModule, 'readPackageManifest')
         .mockResolvedValue({
@@ -118,7 +118,7 @@ describe('package', () => {
       });
     });
 
-    it("returns the fact that the package has not been changed since its latest release if a tag matching the current version exists, and changes have not been made to the package's directory since the tag", async () => {
+    it("does not flag the package as having been changed since its latest release if a tag matching the current version exists, but changes have not been made to the package's directory since the tag", async () => {
       jest
         .spyOn(packageManifestModule, 'readPackageManifest')
         .mockResolvedValue({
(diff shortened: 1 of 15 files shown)


mcmire:
Rebased!

mcmire:
This has been rebased again.

src/package.ts
```diff
@@ -30,28 +32,147 @@ export interface Package {
...
     await readPackageManifest(manifestPath);
+  const expectedReleaseTagNames = generateMonorepoRootPackageReleaseTagNames(
+    validatedManifest.version.toString(),
+  );
+  const tagNameMatchingLatestRelease = expectedReleaseTagNames.find(
+    (tagName) => {
+      return projectTagNames.includes(tagName);
+    },
+  );
+  // TODO: What if tags exist (projectTagNames.length > 0) but there is no tag
```
mcmire, line 98: @Gudahtt What do you think would be a good way to handle this?

src/package.ts
```diff
@@ -30,28 +32,147 @@ export interface Package {
...
+      validatedManifest.name,
+      validatedManifest.version.toString(),
+      rootPackageVersion.toString(),
+    );
+  const tagNameMatchingLatestRelease = expectedReleaseTagNames.find(
+    (tagName) => {
+      return projectTagNames.includes(tagName);
+    },
+  );
+  // TODO: What if tags exist (projectTagNames.length > 0) but there is no tag
```
mcmire, line 157: @Gudahtt What do you think would be a good way to handle this?

src/package.ts
Gudahtt: Hmm. Good question. Maybe it would be best to throw an error in this case? It would mean either that:

* The repository is in an invalid state (missing a tag), which would have to be fixed manually
* Someone is using this tool incorrectly (e.g. manually bumping the manifest version)
* Someone is running this in a repo that doesn't have any tags (e.g. they checked out a commit directly rather than the whole repo).

In all of those cases, we would not want to proceed

Gudahtt: Well, one other case where this would happen is on first run. We do want it to work then. Hmmm.

For the auto-changelog CLI, we assumed that the package had changed if we couldn't find the most recent tag. Maybe that's fine here as well.

src/package.ts
Gudahtt: I think the answer in [my other comment](https://github.com/MetaMask/create-release-branch/pull/23#discussion_r978117112) applies here as well.

src/package.ts
Gudahtt: Maybe we could add a flag at some point to skip tag validation, for if this tool is being first used in a project that used different conventions previously

Gudahtt: Or we could ask the user for confirmation before proceeding

src/repo.ts
```diff
@@ -88,23 +175,96 @@ export async function getRepositoryHttpsUrl(
...
+  const unsortedMergedTagNames = await getLinesFromGitCommandWithin(
+    repositoryDirectoryPath,
+    'tag',
+    ['--merged'],
+  );
+  const sortedComprehensiveTagNames = await getLinesFromGitCommandWithin(
+    repositoryDirectoryPath,
+    'rev-list',
+    ['--tags', '--date-order'],
+  );
```
Gudahtt, line 222: So at first I was going to say that we can get the tags sorted by date with a single command. The `git tag` command accepts a `--sort` parameter that can be used to sort by date.

But, do we want to sort by date here? It seems that the sort order of this list isn't used at all. And even if it was, it would seem more useful to me to sort by semantic order (i.e. `--sort version:refname`) so that backports don't get placed out of order.

src/repo.ts
```diff
@@ -88,23 +175,96 @@ export async function getRepositoryHttpsUrl(
...
+  subdirectoryPath: string,
+  tagName: string,
+): Promise<boolean> {
+  if (!(tagName in CHANGED_FILE_PATHS_BY_TAG_NAME)) {
+    const changedFilePaths = await getFilesChangedSince(
+      repositoryDirectoryPath,
+      tagName,
+    );
+
+    if (!(tagName in CHANGED_FILE_PATHS_BY_TAG_NAME)) {
```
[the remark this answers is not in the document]
Gudahtt: Nevermind, this is required by `require-atomic-updates`

src/repo.ts
mcmire: Yeah, I'm not really using the sorting right now, you're right. I can take the sorting away for now.

src/package.ts
mcmire: > Maybe we could add a flag at some point to skip tag validation, for if this tool is being first used in a project that used different conventions previously

That's a good point. The only new tags we are introducing are tags for monorepo workspace packages. For existing projects, we won't have these tags, and they will only appear after the first release. I have (hopefully) accounted for this by checking for a tag to exist that's formatted like "<package name>@<version>", but *also* checking for a tag that corresponds to the root package, as a fallback.

src/repo.ts
mcmire: Or actually, I will keep the sorting for debugging purposes (in case we need to see what it thinks a repo's tags are) but use the sorting order you recommended.

Gudahtt: If we're not using the order, chronological is fine too, but it would be nice to keep it as one command

src/package.ts
mcmire: Updated here: [4986554](https://github.com/MetaMask/create-release-branch/pull/23/commits/498655440ed6433c5731ad7cb52879ee0b9b7dad)

src/repo.ts
mcmire: Updated here: [5c13376](https://github.com/MetaMask/create-release-branch/pull/23/commits/5c13376989bffe01a527af02f51b746743f07f3c)

src/repo.ts
```diff
@@ -88,23 +174,90 @@ export async function getRepositoryHttpsUrl(
...
+  await getStdoutFromGitCommandWithin(repositoryDirectoryPath, 'add', ['-A']);
+  await getStdoutFromGitCommandWithin(repositoryDirectoryPath, 'commit', [
     '-m',
     `Release ${releaseVersion}`,
   ]);
 }
+
+/**
+ * Retrieves the names of the tags in the given repo, sorted by earliest created
+ * to latest created. As this fetches tags from the remote first, you are
```
Gudahtt, line 198: Nit: Update description to match change in sort order

```suggestion
 * Retrieves the names of the tags in the given repo, sorted by tag name in
 * ascending order. As this fetches tags from the remote first, you are
```

src/repo.test.ts
```diff
@@ -117,4 +103,125 @@ describe('git-utils', () => {
       );
     });
   });
+
+  describe('getTagNames', () => {
+    it('returns all of the tag names that match a known format, sorted from earliest created to latest created', async () => {
```
Gudahtt, line 108: Nit: Update description to match change in sort order
```suggestion
    it('returns all of the tag names that match a known format, sorted by tag name in ascending order', async () => {
```

src/project.ts
```diff
@@ -79,7 +84,12 @@ export async function readProject(
   projectDirectoryPath: string,
 ): Promise<Project> {
   const repositoryUrl = await getRepositoryHttpsUrl(projectDirectoryPath);
-  const rootPackage = await readPackage(projectDirectoryPath);
+  const tagNames = await getTagNames(projectDirectoryPath);
+  const rootPackage = await readMonorepoRootPackage({
```
Gudahtt, line 88: Nit: In this context, this "monorepo root package" might be a standalone package, rather than the root of a monorepo. Maybe we should drop "monorepo" from this function name so that isn't misunderstood.

Gudahtt: Oh wait, nevermind, this is only for the monorepo case. So maybe it's the description of this function that should change

src/package.ts
```diff
@@ -30,28 +33,181 @@ export interface Package {
...
+  const expectedWorkspacePackageReleaseTagNames =
+    generateMonorepoWorkspacePackageReleaseTagNames(
+      validatedManifest.name,
+      validatedManifest.version.toString(),
+    );
+  const expectedRootPackageReleaseTagNames =
+    generateMonorepoRootPackageReleaseTagNames(rootPackageVersion.toString());
+  const tagNameMatchingLatestRelease = [
+    ...expectedWorkspacePackageReleaseTagNames,
+    ...expectedRootPackageReleaseTagNames,
```
Gudahtt, line 169: Perhaps it's worth printing a console warning if the fallback root tag is used? This should only happen in rare circumstances, like when using this tool for the first time on a monorepo that didn't previously have workspace tags (or at least not in the expected format).

A warning might hint that something is amiss if this happens outside of that scenario.

src/package.test.ts
```diff
@@ -2,40 +2,426 @@ import fs from 'fs';
...
+        projectDirectoryPath: '/path/to/project',
+        projectTagNames: ['v1.0.0'],
+      });
+
+      expect(pkg).toMatchObject({
+        hasChangesSinceLatestRelease: false,
+      });
+    });
+
+    it("returns the fact that the package has been changed since its latest release, if a tag matching 'v' + the current version exists and changes have been made to the package's directory since the tag", async () => {
```
Gudahtt, line 145: This test looks identical to the one two before it :thinking:

src/repo.ts
mcmire: Good catch. Maybe for clarity we say "ascending semantic version order"?

src/project.ts
mcmire: Just to be clear you're saying I should drop mention of "polyrepo" from the description for now, yeah?

src/package.ts
mcmire: If we display a message to the user, I'd like it to be helpful and include a suggestion as to how they can resolve the problem, but I'm curious what that suggestion would be, because I think it differs based on whether this is the first time this tool is being used on a monorepo or not. If it is not the first time, then we can presume that all packages have been tagged properly, and so if the current release for any given package does _not_ have the tag we expect, that is an anomaly and should be corrected. However, if it _is_ the first time, then we can presume that no releases for any of the packages in the monorepo have tags at all, so in that case, if we say that the current release for a given package needs to be tagged properly, we might as well say that ALL releases for a given package needs to be tagged properly.

So, either we need a way to tell whether this is the first time a user is using this tool, or we need to offer another tool the user can use which will validate that all releases for all packages have proper tags and backfill them if necessary.

Thoughts/

src/project.ts
Gudahtt: Yeah exactly; the part about how it handles polyrepo packages isn't accurate anymore because of these changes, so it can be dropped

src/repo.ts
Gudahtt: Sounds reasonable!

src/package.ts
Gudahtt: Originally I was thinking the user would just want to know the fallback occurred, since as you point out it's not necessarily a problem or something that is actionable.

If we can do better, that would be great. I'll give this some thought.

src/package.test.ts
mcmire: Ah. I only spotted one duplicate, so I removed that. As for the other tests, they follow a "this should be true if this is the case" / "this should not be true if this is not the case" pattern, so perhaps that was difficult to see. I renamed the tests to make that distinction more clear.

src/project.ts
mcmire: Fixed in [8f7703f](https://github.com/MetaMask/create-release-branch/pull/23/commits/8f7703f6ab475d5368f0b4404866163e13aac6f8).

src/package.test.ts
```diff
@@ -402,7 +354,36 @@ describe('package', () => {
       });
     });
 
-    it("returns the fact that the package has been changed since its latest release if a tag matching neither the package name + version nor 'v' + the root package version exists", async () => {
+    it("prints a warning if a tag matching 'v' + the root package version exists instead of the package name + version", async () => {
```
mcmire: This doesn't belong here yet :)

src/repo.test.ts
mcmire: Fixed in [8190f5f](https://github.com/MetaMask/create-release-branch/pull/23/commits/8190f5f0a7cc46ec85269ed8645c909df555e757).

src/repo.ts
mcmire: Fixed in [8190f5f](https://github.com/MetaMask/create-release-branch/pull/23/commits/8190f5f0a7cc46ec85269ed8645c909df555e757).

src/package.ts
Gudahtt: >If it is not the first time, then we can presume that all packages have been tagged properly, and so if the current release for any given package does not have the tag we expect, that is an anomaly and should be corrected.

Yes, agreed. We could even throw an error in this case, this should never happen. That would be an improvement. It's distinct from the suggestion I meant to make here though.

>However, if it is the first time, then we can presume that no releases for any of the packages in the monorepo have tags at all, so in that case, if we say that the current release for a given package needs to be tagged properly, we might as well say that ALL releases for a given package need to be tagged properly.

If we're just starting to use this tool, I don't think we can make any assumptions about tags. We don't know what convention, if any, they were using before.

To be clear, I was suggesting that we warn in the case where we find a root monorepo tag but not a workspace tag. We proceed silently in that circumstance in this PR at the moment.

This seems worth highlighting so that the user understands it's happening, even just for auditability. For example, if there were unpublished changes in a workspace that predate the most recent root tag, they would be missed because of us using this fallback. I don't think this scenario is likely, nor do I think it makes sense necessarily to force the user to manually add tags. But calling out that it's happening might leave an audit trail to follow up on if it happens and someone is confused as to why.

Gudahtt (dismissed):
LGTM! one suggestion thread pending but it's non-blocking

src/package.test.ts
Gudahtt: Huh, it seems that both duplicate tests I saw are removed in that commit. The re-wording looks good too though!

src/package.ts
mcmire: Fair enough! Added warning in [565c2a5](https://github.com/MetaMask/create-release-branch/pull/23/commits/565c2a505879520c811bf593e75a42913590f313).

Gudahtt (approved):
LGTM!

PR #27, opened 2022-08-02
Bump @metamask/utils to 2.1.0

mcmire:
Supercedes #26 (see there for release notes). This update has no impact on this tool but we might as well bump it.


mcmire:
Will bump this later, now that 3.0.1 is out.

PR #29, opened 2022-08-04
Discourage postponing release of changed packages

### LINKED ISSUE #10: Monorepo flow: Warn the user if any packages in the release spec fall outside the set of changed packages
When editing the release spec, the user technically has the opportunity to modify the set of packages listed there, but this should be discouraged as much as possible, because it means that it is possible to release a package that relies upon another package which is not being published. For instance, say B is bumped and A is bumped but only B is released. This may work just fine in development and on CI but will *not* work in production. We want to prevent this from happening.

mcmire:
When the tool generates a release spec, it will list all packages that
have changed since their last release. Editing the release spec to
remove a package so that it is no longer included in the release is
potentially dangerous, because it means that any package which relies on
that package that *is* included in the release could be broken in
production.

Since it's sometimes necessary to delay the release of a package, this
commit changes the release spec validation step such that if it detects
that a package that should be listed in the release spec *isn't*, it
will throw an error, advising the user of the danger, yet provide a
hidden option should the user really want to proceed.

---

Fixes #10.

### DIFF
--- a/src/functional.test.ts
+++ b/src/functional.test.ts
@@ -32,11 +32,6 @@ describe('create-release-branch (functional)', () => {
               version: '1.2.3',
               directoryPath: 'packages/d',
             },
-            e: {
-              name: '@scope/e',
-              version: '0.0.3',
-              directoryPath: 'packages/e',
-            },
           },
           workspaces: {
             '.': ['packages/*'],
@@ -64,11 +59,6 @@ describe('create-release-branch (functional)', () => {
             },
           });
           await environment.updateJsonFileWithinPackage('d', 'package.json', {
-            scripts: {
-              foo: 'bar',
-            },
-          });
-          await environment.updateJsonFileWithinPackage('e', 'package.json', {
             scripts: {
               foo: 'bar',
             },
@@ -118,13 +108,6 @@ describe('create-release-branch (functional)', () => {
           ).toStrictEqual({
             name: '@scope/d',
             version: '1.2.4',
-            scripts: { foo: 'bar' },
-          });
-          expect(
-            await environment.readJsonFileWithinPackage('e', 'package.json'),
-          ).toStrictEqual({
-            name: '@scope/e',
-            version: '0.0.3',
             scripts: { foo: 'bar' },
           });
         },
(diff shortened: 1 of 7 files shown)


mcmire:
Rebased!

mcmire:
Rebased again.

src/functional.test.ts
```diff
@@ -32,11 +32,6 @@ describe('create-release-branch (functional)', () => {
               version: '1.2.3',
               directoryPath: 'packages/d',
             },
-            e: {
```
Gudahtt, line 35: The reason for this removal isn't clear to me, but I can't see why this would be needed either. Just tidying up? Or is there something I'm missing.

src/functional.test.ts
```diff
@@ -289,5 +272,115 @@ describe('create-release-branch (functional)', () => {
...
+
+  - @scope/b
+  - @scope/d
+
+  Consider including them in the release spec so that any packages that rely on them won't break in production.
+
+  If you are ABSOLUTELY SURE that this won't occur, however, and want to postpone the release of a package, then list it with a directive of "intentionally-skip". For example:
+
+    packages:
+      "@scope/b": intentionally-skip
```
Gudahtt, line 334: Could you add a test to cover the `intentionally-skip` flag as well?

tests/setupAfterEnv.ts
```diff
@@ -79,4 +90,51 @@ expect.extend({
...
+      return {
+        message: () =>
+          'Expected running the tool to fail with the given error message, but it did not.',
+        pass: false,
+      };
+    } catch (error) {
+      if (isExecaError(error)) {
+        const stderr = [
+          {
+            from: /^\s+at.+\)$/msu,
```
Gudahtt, line 110: Nit: A comment explaining what we expect this to match might help! I'm not familiar with the execa error output.

Gudahtt (approved):
LGTM! Just a couple of questions, and one suggested additional test.

Gudahtt:
The hidden "intentionally-skip" flag seems like a great idea. It's well documented, but only in the case you run into it. Good compromise for something we want to discourage.

src/functional.test.ts
mcmire: These function tests started out as tests I'd written for `action-create-release-pr`, and if I remember correctly, I'd added a package that had a 0.x version because that action has special behavior around 0.x releases. We don't have that in this case, so I figured that this was extraneous.

src/functional.test.ts
mcmire: Good catch. Added here: [6e94e30](https://github.com/MetaMask/create-release-branch/pull/29/commits/6e94e30f40cbe1d5f6106b500ff7d9d4f23834ea)

tests/setupAfterEnv.ts
mcmire: I've extracted this to a constant and added some docs to explain what it is. I also realized that this should have been "stack trace" rather than "backtrace" (I've been writing too much Ruby lately, apparently): 6e94e30...ad4fe11

src/functional.test.ts
```diff
@@ -289,5 +272,289 @@ describe('create-release-branch (functional)', () => {
...
+            await environment.readJsonFileWithinPackage('d', 'package.json'),
+          ).toStrictEqual({
+            name: '@scope/d',
+            version: '1.2.3',
+          });
+        },
+      );
+    });
+
+    it('does not update the changelogs of any packages that have been tagged with --intentionally-skip', async () => {
```
Gudahtt, line 466: Nit: I spent a while comparing this with the previous test, trying to see what the difference between `intentionally-skip` and `--intentionally-skip` was. I'm guessing it was a typo? This normalizes the two test descriptions

```suggestion
    it('does not update the changelogs of any packages that have been tagged with intentionally-skip', async () => {
```

mcmire: Yes, thank you!

PR #39, opened 2022-10-10
Fix hasChangesInDirectorySinceGitTag

mcmire:
To determine whether a package needs a new release, we ask if there have been any new commits that contain changes to any of the files within that package. In other words, we run a `git diff` on the repo and look for any files that match that package.

The problem is that `git diff` returns partial file paths, that is, file paths which are relative to the repository directory, whereas the subdirectory path that `hasChangesInDirectorySinceGitTag` takes is an absolute path. So when we look for matching packages we always come up short. This confuses the tool because it doesn't think there are any new changes to release.

### DIFF
--- a/src/repo.test.ts
+++ b/src/repo.test.ts
@@ -175,10 +175,7 @@ describe('repo', () => {
         .calledWith('git', ['diff', 'v1.0.0', 'HEAD', '--name-only'], {
           cwd: '/path/to/repo',
         })
-        .mockResolvedValue([
-          '/path/to/repo/file1',
-          '/path/to/repo/subdirectory/file1',
-        ]);
+        .mockResolvedValue(['file1', 'subdirectory/file1']);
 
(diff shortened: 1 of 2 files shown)


Gudahtt (approved):
LGTM!

Though, have you considered using `git diff --quiet [repo-path]` instead? That could give us a status (has the directory changed, yes or no) for each package, rather than a list of files. Might be easier to manage.

mcmire:
@Gudahtt Ah, so I think the reason why this command pulls a list of files rather than a status is so that if we want to know whether multiple packages changed but we are using the same tag to get this information, then the command is run only once (because `git diff <tag>` is cached) rather than multiple times (because `git diff <tag> <path1>` would get run, then `git diff <tag> <path2>`, etc.)

Gudahtt:
The full list of changed files could be quite large, even if we're only asking for it once. I would expect getting a per-package status would be similarly fast or faster in all cases. But that's fine, this works for now.

PR #40, opened 2022-10-11
Update docs to reflect recent changes

mcmire:
* Document backport releases
* Correct format of root version and explanation for how it gets incremented


Gudahtt (approved):
LGTM!

PR #42, opened 2022-10-12
Fix published executable

mcmire:
* Add a hashbang line to ensure that it will be executed with Node as per the [instructions in the NPM docs][1]
* Ensure that the name of the executable is `create-release-branch`, not `@metamask/create-release-branch`

[1]: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#bin

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,18 +6,20 @@
     "type": "git",
     "url": "https://github.com/MetaMask/create-release-branch.git"
   },
-  "bin": "dist/cli.js",
+  "bin": {
+    "create-release-branch": "dist/cli.js"
+  },
   "files": [
     "dist/"
   ],
   "scripts": {
-    "build": "tsc --project tsconfig.build.json",
-    "build:clean": "rimraf dist && yarn build",
+    "build": "./scripts/build.sh",
+    "build:clean": "rimraf dist tmp && yarn build",
     "lint": "yarn lint:eslint && yarn lint:misc --check",
     "lint:eslint": "eslint . --cache --ext js,ts",
     "lint:fix": "yarn lint:eslint --fix && yarn lint:misc --write",
     "lint:misc": "prettier '**/*.json' '**/*.md' '!CHANGELOG.md' '**/*.yml' '!.yarnrc.yml' --ignore-path .gitignore --no-error-on-unmatched-pattern",
-    "prepack": "yarn build:clean && chmod +x dist/cli.js && yarn lint && yarn test",
+    "prepack": "yarn lint && yarn build:clean && yarn test",
     "test": "jest && jest-it-up",
     "test:watch": "jest --watch"
   },
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+yarn tsc --project tsconfig.build.json
+mkdir tmp
+touch tmp/cli.js
+echo "#!/usr/bin/env node" >> tmp/cli.js
+echo >> tmp/cli.js
+cat dist/cli.js >> tmp/cli.js
+mv tmp/cli.js dist/cli.js
+rm -r tmp
+chmod +x dist/cli.js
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -6,18 +6,20 @@
     "type": "git",
     "url": "https://github.com/MetaMask/create-release-branch.git"
   },
-  "bin": "dist/cli.js",
+  "bin": {
+    "create-release-branch": "dist/cli.js"
```
Gudahtt, line 10: Not sure this part is necessary - npm should automatically make it available without the organization prefix.

scripts/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+yarn tsc --project tsconfig.build.json
+mkdir tmp
+touch tmp/cli.js
+echo "#!/usr/bin/env node" >> tmp/cli.js
```
Gudahtt, line 8: Rather than dynamically add the shebang, can we add it directly to `cli.js`?

package.json
mcmire: Are we sure about that? The docs say:

> If you have a single executable, and its name should be the name of the package, then you can just supply it as a string.

But the name of the executable is not the name of the package in this case.

scripts/build.sh
mcmire: I thought about adding it to `cli.ts`, but I thought that would be confusing, because that file can't be run through Node, it has to be run through `ts-node` instead.

package.json
Gudahtt: I guess the docs don't account for package namespaces. In practice the namespace does not get included in the binary name. You can check for yourself in `node_modules/.bin`. `auto-changelog` works this way for example, the package being called `@metamask/auto-changelog`.

Gudahtt: I mean, I guess it couldn't hurt to explicitly name the binary :shrug: probably.

mcmire: Ah, right. I keep forgetting we already have a package that does this. Interesting, okay.

scripts/build.sh
Gudahtt: Ah I see. Makes sense. Hmm. Maybe we could use a wrapper Node.js script instead? e.g.

```javascript
#!/usr/bin/env node

require('./dist/cli.js')()
```

That would also remove the need for a build script.

package.json
mcmire: Yeah... I guess the alternative is to update the NPM documentation, but absent of that, I can imagine myself getting confused about this again in the future.

scripts/build.sh
mcmire: Okay, do you propose still building this on the fly in `scripts/build.sh` or putting this in `src/`?

Gudahtt: Oh I guess it could go in `src`. Another pattern I've seen before is creating a `bin` directory for the wrapper script. Either way, it avoids the need to move anything around at build time.

package.json
Gudahtt: ... is our linter objecting to this? :thinking:

mcmire: For some reason, `package.json` gets changed while running the action. I'm not totally sure why...

mcmire: Oh, `yarn install` seems to change it. Well then!

Gudahtt (approved):
LGTM!

package.json
```diff
@@ -6,18 +6,19 @@
...
   },
-  "bin": "dist/cli.js",
+  "bin": "bin/create-release-branch.js",
   "files": [
+    "bin/",
     "dist/"
   ],
   "scripts": {
-    "build": "tsc --project tsconfig.build.json",
+    "build": "yarn tsc --project tsconfig.build.json",
```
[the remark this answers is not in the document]
Gudahtt: Oh, you got there first!

package.json
mcmire: This is now simplified again.

scripts/build.sh
mcmire: I've added a script to `bin/`.

PR #44, opened 2022-11-04
Update GitHub actions to match module template

Gudahtt:
The GitHub actions have been updated to match the module template, with some exceptions. The docs publishing was omitted because this repository does not yet have API documentation, and I left some additional steps in the prepack script because they were present already.


bin/create-release-branch.js
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env node
 
-/* eslint-disable-next-line import/no-unassigned-import */
+/* eslint-disable-next-line import/no-unassigned-import,import/no-unresolved */
```
Gudahtt, line 3: I had to ignore `import/no-unresolved` here to enable running the linter in a parallel job to the build. It only passed the linter before because the build ran first.

Gudahtt:
This depends upon #45(for the lint error) and #46 (for the Node.js v19 error)

rickycodes (approved):

PR #47, opened 2022-11-07
Update TypeScript to the latest stable version

Gudahtt:
TypeScript has been updated to the latest stable version. Additional TypeScript options added in these versions have been migrated into the `tsconfig.json` from the MetaMask module template.

One option was kept back for now: `noUncheckedIndexedAccess`. This option resulted in too many errors, so will be attempted in a separate PR.

The range was updated to use `~` because TypeScript doesn't follow SemVer; their minor releases can include breaking changes.


Gudahtt:
Issue created here to track adding `noUncheckedIndexedAccess`: https://github.com/MetaMask/create-release-branch/issues/49

rickycodes (approved):

PR #48, opened 2022-11-07
Update ESLint config from v9 to v10

Gudahtt:
The ESLint configuration has been updated from v9 to v10, and all related packages has been updated. This resolves the console warning that had been printed upon each run of `yarn lint` about how the current TypeScript version is unsupported.

All lint changes were made with `yarn lint:fix` except one, which is where we're using `interface` over `type` to allow for declaration merging in `setupAfterEnv.ts`.


rickycodes (dismissed):

Gudahtt:
Rebased to resolve conflicts

rickycodes (approved):

PR #66, opened 2022-12-19
Remove warning from README, and tighten summary

mcmire:
This tool is actively being used to create new releases for the `controllers` monorepo, and we welcome people to use it for other projects.


Gudahtt:
Should we leave in a warning about this only supporting independently versioned monorepos?

mcmire:
@Gudahtt Ah. I suppose you're right, I guess we still need some kind of message, even if it's more of a disclaimer rather than an outright warning.

Gudahtt (approved):
LGTM!

PR #79, opened 2023-04-26
chore: bump dependencies

legobeat:
https://github.com/MetaMask/create-release-branch/security/dependabot/6

### DIFF
--- a/src/functional.test.ts
+++ b/src/functional.test.ts
@@ -654,16 +654,16 @@ Error: Your release spec could not be processed due to the following issues:
 
 * The following packages, which have changed since their latest release, are missing.
 
-  - @scope/b
   - @scope/d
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "moduleResolution": "node",
     "noEmit": true,
     "noErrorTruncation": true,
+    "skipLibCheck": true,
```
Gudahtt, line 11: Hmm. It's a shame to disable type checking for libraries. I see that you found [an error that there is no apparently solution for](https://github.com/isaacs/node-lru-cache/issues/292#issuecomment-1524256758) though. I guess we can use this as a workaround in the meantime.

Gudahtt (approved):
LGTM!

PR #83, opened 2023-07-14
Add `.yml` extension to `RELEASE_SPEC` file

Mrtenz:
The release spec file is a YAML file, but right now does not have an extension. I've added an extension to the file to ensure that editors supporting YAML read the file as such, mainly so it adds syntax highlighting.


mcmire (approved):
Makes sense!

PR #84, opened 2023-07-14
Add support for nested workspaces

Mrtenz:
This adds support for nested workspaces (workspaces in workspaces), by updating the logic to fetch workspaces. Rather than looking at the root `package.json`, it now runs `yarn workspaces list --no-private` to get a list of all workspaces.

### DIFF
--- a/src/functional.test.ts
+++ b/src/functional.test.ts
@@ -1,7 +1,7 @@
 import { withMonorepoProjectEnvironment } from '../tests/functional/helpers/with';
 import { buildChangelog } from '../tests/functional/helpers/utils';
 
-jest.setTimeout(60_000);
+jest.setTimeout(10_000);
 
 describe('create-release-branch (functional)', () => {
   describe('against a monorepo with independent versions', () => {
--- a/src/functional.test.ts
+++ b/src/functional.test.ts
@@ -1,6 +1,8 @@
 import { withMonorepoProjectEnvironment } from '../tests/functional/helpers/with';
 import { buildChangelog } from '../tests/functional/helpers/utils';
 
+jest.setTimeout(60_000);
+
 describe('create-release-branch (functional)', () => {
   describe('against a monorepo with independent versions', () => {
     it('bumps the ordinary part of the root package and updates the versions of the specified packages according to the release spec', async () => {
@@ -654,16 +656,16 @@ Error: Your release spec could not be processed due to the following issues:
 
 * The following packages, which have changed since their latest release, are missing.
 
-  - @scope/d
   - @scope/b
+  - @scope/d
 
   Consider including them in the release spec so that any packages that rely on them won't break in production.
 
   If you are ABSOLUTELY SURE that this won't occur, however, and want to postpone the release of a package, then list it with a directive of "intentionally-skip". For example:
 
     packages:
-      "@scope/d": intentionally-skip
       "@scope/b": intentionally-skip
+      "@scope/d": intentionally-skip
 
 The release spec file has been retained for you to edit again and make the necessary fixes. Once you've done this, re-run this tool.
 
--- a/src/misc-utils.ts
+++ b/src/misc-utils.ts
@@ -155,7 +155,7 @@ export async function getStdoutFromCommand(
 }
 
 /**
- * Runs a Git command, splitting up the immediate output into lines.
+ * Run a command, splitting up the immediate output into lines.
  *
  * @param command - The command to execute.
  * @param args - The positional arguments to the command.
--- a/src/project.ts
+++ b/src/project.ts
@@ -8,7 +8,6 @@ import {
 } from './package';
 import { getRepositoryHttpsUrl, getTagNames } from './repo';
 import { SemVer } from './semver';
-import { getLinesFromCommand } from './misc-utils';
 import { PackageManifestFieldNames } from './package-manifest';
 
 /**
--- a/src/project.ts
+++ b/src/project.ts
@@ -77,13 +77,15 @@ function examineReleaseVersion(packageVersion: SemVer): ReleaseVersion {
 export async function listWorkspaces(projectDirectoryPath: string) {
   const stdout = await getLinesFromCommand(
     'yarn',
-    ['workspaces', 'list', '--no-private', '--json'],
+    ['workspaces', 'list', '--json'],
     {
       cwd: projectDirectoryPath,
     },
   );
 
-  return stdout.map((line) => JSON.parse(line) as YarnWorkspace);
+  return stdout
+    .map((line) => JSON.parse(line) as YarnWorkspace)
+    .filter(({ location }) => location !== '.');
 }
 
 /**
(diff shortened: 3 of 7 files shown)


src/misc-utils.ts
```diff
@@ -155,7 +155,7 @@ export async function getStdoutFromCommand(
 }
 
 /**
- * Runs a Git command, splitting up the immediate output into lines.
+ * Run a command, splitting up the immediate output into lines.
```
Mrtenz, line 158: This function can run any command, not just Git commands.

src/project.ts
```diff
@@ -63,6 +68,24 @@ function examineReleaseVersion(packageVersion: SemVer): ReleaseVersion {
...
+/**
+ * List the workspaces in a monorepo using Yarn.
+ *
+ * @param projectDirectoryPath - The path to the project.
+ * @returns An array of objects that represent the workspaces.
+ */
+export async function listWorkspaces(projectDirectoryPath: string) {
+  const stdout = await getLinesFromCommand(
+    'yarn',
+    ['workspaces', 'list', '--no-private', '--json'],
```
Mrtenz, line 80: Note that I've added `--no-private`. I'm not 100% sure if that's expected, but works for the Snaps repository at least.

tests/functional/helpers/local-monorepo.ts
```diff
@@ -50,6 +50,12 @@ export default class LocalMonorepo<
     this.#workspaces = workspaces;
   }
 
+  async create() {
+    await super.create();
+    await this.runCommand('yarn', ['set', 'version', 'stable']);
+    await this.runCommand('yarn');
+  }
```
Mrtenz, line 57: `yarn workspaces list` requires Yarn 3. I'm not sure if there's a better way to handle this.

src/project.ts
Mrtenz: Actually, we still want to version those. 🤔 I have to re-think this.

tests/functional/helpers/local-monorepo.ts
mcmire: True, this change would make this tool usable only on Yarn v3 projects. I am okay with this, but it is inconsistent with recent decisions, so it could create possible confusion if we want to expand the scope of this tool.

A bit of backstory: When we converted the `core` repo to a monorepo we decided to follow after the Snaps monorepo and use Yarn v3. After that, we effectively decided that Yarn v3 was the future and updated `action-publish-release` and `action-npm-publish` to only support v3. Libraries that were using v1 could continue use older versions of these actions and eventually once they supported v3, everything would be fine.

However, it sounds like, based on some recent changes, this plan may not be as practical as we'd hoped, because I see that we've effectively restored Yarn v1 support for polyrepos to `action-publish-release` (https://github.com/MetaMask/action-publish-release/pull/83) and `action-npm-publish` (https://github.com/MetaMask/action-npm-publish/pull/51/files).

So that leaves the question of what to do here. On one hand, `create-release-branch` is only used for `core` at the moment, which does support Yarn v3, and if we want to use this for Snaps too, then that would work fine since it uses v3 as well. So, one route is to change this tool to explicitly support v3 only.

However, if anyone wanted to extend this tool to support polyrepo packages, we would be back in the same boat we're in now, and we'd have to come up with a Yarn v1-compatible solution. Technically we already have one: [`getWorkspaceLocations` in `action-utils`](https://github.com/MetaMask/action-utils/blob/b33c0b6b5c6cb3d544f81231f6187a2925841ff5/src/package-utils.ts#L280). So, we could use this implementation instead of or even as an alternative to `yarn workspaces list`. 

What is your opinion?

Mrtenz: > A bit of backstory: When we converted the core repo to a monorepo we decided to follow after the Snaps monorepo and use Yarn v3. After that, we effectively decided that Yarn v3 was the future and updated action-publish-release and action-npm-publish to only support v3. Libraries that were using v1 could continue use older versions of these actions and eventually once they supported v3, everything would be fine.

> However, it sounds like, based on some recent changes, this plan may not be as practical as we'd hoped, because I see that we've effectively restored Yarn v1 support for polyrepos to action-publish-release (https://github.com/MetaMask/action-publish-release/pull/83) and action-npm-publish (https://github.com/MetaMask/action-npm-publish/pull/51). (Unless I'm misunderstanding the intent of these PRs?)

For these PRs I assumed we want to keep supporting Yarn v1 for the time being, since some of our polyrepos use it. In MetaMask/action-publish-release#82 I introduced `yarn npm info`, but later realised this doesn't work for Yarn v1, so I changed it back in MetaMask/action-publish-release#83 to fix it. For the implementation in MetaMask/action-npm-publish#51 I simply copied the other implementation. The goal of those PRs was to support tags (since we release some packages in the Snaps repo under the `flask` tag), not to support Yarn v1 per se.

If we already broke Yarn v1 compatibility for these packages (I'm not sure if we did?) I suppose the PR to fix compatibility wasn't necessary. I didn't see anything that indicated Yarn v3 support only though. 

> However, if anyone wanted to extend this tool to support polyrepo packages, we would be back in the same boat we're in now, and we'd have to come up with a Yarn v1-compatible solution. In fact I believe we may already have one: [getWorkspaceLocations in action-utils](https://github.com/MetaMask/action-utils/blob/b33c0b6b5c6cb3d544f81231f6187a2925841ff5/src/package-utils.ts#L280). So, we could use this implementation instead of or even as an alternative to yarn workspaces list.

I hadn't thought about using this function. That seems like a better alternative!

Mrtenz:
@SocketSecurity ignore @metamask/action-utils@1.0.0

This is our own.

src/functional.test.ts
```diff
@@ -1,6 +1,8 @@
 import { withMonorepoProjectEnvironment } from '../tests/functional/helpers/with';
 import { buildChangelog } from '../tests/functional/helpers/utils';
 
+jest.setTimeout(60_000);
```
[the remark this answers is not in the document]
Mrtenz: The tests were a bit flaky, but you're right that 60 seconds is not necessary. I'll change it to 10 seconds.

src/project.ts
```diff
@@ -27,6 +28,11 @@ type ReleaseVersion = {
   backportNumber: number;
 };
 
+export type YarnWorkspace = {
```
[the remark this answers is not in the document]
Mrtenz: Not anymore. Removed it now.

src/functional.test.ts
```diff
@@ -1,6 +1,8 @@
 import { withMonorepoProjectEnvironment } from '../tests/functional/helpers/with';
 import { buildChangelog } from '../tests/functional/helpers/utils';
 
+jest.setTimeout(10_000);
+
 describe('create-release-branch (functional)', () => {
```
mcmire, line 6: It would be nice if we could write a functional test involving nested workspaces, but we can do that later.

mcmire (approved):
Looks good!

PR #87, opened 2023-07-19
Correct setup instructions for a monorepo w/ independent versions

mcmire:
The setup instructions for a monorepo with independently released versions mentions that the version of the root package needs to be formatted a certain way. This was true when we were first developing this tool, but we have changed this requirement since.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -35,7 +35,6 @@ Combined with the changes above, this might look like:
 
 For a monorepo with independent versions, you will want to add a `release.config.json` with a `versioningStrategy` of `"independent"`.
 
-
 Combined with the changes above, this might look like:
 
 ```
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -33,17 +33,15 @@ Combined with the changes above, this might look like:
 
 ### Monorepos with independent versions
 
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -33,17 +33,15 @@ Combined with the changes above, this might look like:
 
 ### Monorepos with independent versions
 
-For a monorepo with independent versions, you will want to:
+For a monorepo with independent versions, you will want to add a `release.config.json` with a `versioningStrategy` of `"independent"`.
```
mcmire, line 36: Note that we don't actually use the `release.config.json` right now. In fact this documentation is mostly aspirational and reflects future plans. Perhaps I ought to make that clear in another set of changes, but I wanted to call that out in case it was confusing.

Gudahtt: This file is used by https://github.com/MetaMask/action-publish-release

Gudahtt (dismissed):
LGTM!

docs/setup.md
mcmire: Oh, right, we did decide to make of it there, didn't we. Okay, I'll add some documentation to that action to mention that.

mcmire:
Had a lint violation 🤦🏻

Issue #92, opened 2023-08-15
Remove error shown when omitting changed packages from release spec


mcmire:
In order to support partial releases of a monorepo — that is, the ability to publish a subset of publishable packages within a release — we need to allow users to omit packages from the release spec that have changes since their previous release.

To do this, we need to:

- Remove the check that runs when a version is not specified for a package that is checked
- Remove the error that is thrown when that happens
- Remove `intentionally-skip` as a possible version specifier for the release spec

Part of https://github.com/MetaMask/core/issues/1539.

cryptodev-2s:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Shared-Libraries-621e46b4d7103800171d1b02/issues/MetaMask/create-release-branch/92?planning-poker) with Zenhub @mikesposito @mcmire @Gudahtt

Issue #91, opened 2023-08-15
Compel users to release new versions of dependencies alongside their dependents


mcmire:
Let's say that `controller-utils` relies upon `network-controller`:

``` mermaid
graph BT
  controller-utils --> network-controller
```

Now `controller-utils` introduces a new function that is used by `network-controller`. `network-controller` is then released without `controller-utils` also being released. What would happen? This would break things in the clients when `network-controller` was updated to use the new version, because the new function in `controller-utils` it was attempting to use would not exist.

To address this, we can make a change to this tool.

Given the project is a monorepo and the tool is applying a release spec, when the user has indicated an addition or breaking change to a dependency, B, of a package, A, by bumping B's major or minor version, but the user has not included B in the release spec, then warn the user so that they are compelled to include it in the same way that the tool warns the user about excluding packages with any kinds of changes today: that is, explain the consequences of not including the package and force the user to re-edit the release spec and indicate `intentionally-skip` for B's version if they truly want to bypass the warning.

xinnanyemm:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Shared-Libraries-621e46b4d7103800171d1b02/issues/MetaMask/create-release-branch/91?planning-poker) with Zenhub @cryptodev-2s @Gudahtt @mikesposito

Issue #90, opened 2023-08-15
Compel users to release packages with breaking changes alongside their dependents


mcmire:
Let's say we have the following package tree:

``` mermaid
graph BT
  assets-controller -- dependency --> network-controller
  gas-fee-controller -- peer dependency --> network-controller
```

Let's say that `network-controller` has a breaking change, but it is released without either `assets-controllers` or `gas-fee-controller` being released. What would happen? This would break things in the clients when `network-controller` was updated to use the new version, because `assets-controllers` and `gas-fee-controller` would assume that it was interfacing with the older version of `network-controller` when this was not the case.

To address this, we can make two interrelated changes to this tool:

**Given** the project is a monorepo
**And** the user has indicated a breaking change in the release spec for a package, A, by specifying `major` for the version
**And** another package, B, relies on A either via `peerDependencies` or `dependencies`
**When** the tool applies the release spec
**Then** add an entry to the changelog for B that mentions the version bump so as to ensure that B is included in the release

**Given** the project is a monorepo
**And** the user has indicated a breaking change in the release spec for a package, A, by specifying `major` for the version
**And** another package, B, relies on A either via `peerDependencies` or `dependencies`
**And** the user has not included B in the release spec
**When** the tool applies the release spec
**Then** warn the user so that they are compelled to include in the same way that the tool warns the user about excluding packages with changes today: that is, explain the consequences of not including the package and force the user to re-edit the release spec and indicate `intentionally-skip` for B's version if they truly want to bypass the warning.

xinnanyemm:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Shared-Libraries-621e46b4d7103800171d1b02/issues/MetaMask/create-release-branch/90?planning-poker) with Zenhub @Gudahtt @mikesposito @cryptodev-2s

PR #98, opened 2023-09-12
Allow users to omit packages

cryptodev-2s:
This PR allows users to omit packages from the release spec that have changes since their previous release to make it possible to make partial releases.

Issue: [92](https://github.com/MetaMask/create-release-branch/issues/92)


Gudahtt:
@legobeat Thanks, good idea, I have approved your other PR. But I still think we need something like this PR to make the process for omitting packages easier.

Originally we had internationally added friction to dissuade people from omitting packages because omitting packages can be hazardous in ways that are easy to miss. But #90 and #91 should make the process a lot safer. After those are resolved, we want to make partial releases a low-friction process. Typing `intentionally-skip` 12 times isn't quite low friction enough to make this a pleasant tool to use.

I would suggest putting this on hold until after #90 and #91 are resolved, then merging this or something like it. We can consider alternative approaches to this PR as well, so long as they also achieve the goal of making partial releases low-friction.

mcmire:
Given Mark's comment, does it make sense to mark this ticket as being blocked in some way?

cryptodev-2s:
> Given Mark's comment, does it make sense to mark this ticket as being blocked in some way?

it's currently marked as blocked by 90 and 91 which there's two PRs waiting for review

mcmire:
Okay, I've marked it as such using Zenhub.

cryptodev-2s:
@legobeat PR updated now intentionally-skip is mandatory only when a package is used by another package or uses a package not on the release spec

mcmire (approved):
Looks good to me!

PR #101, opened 2023-09-28
Compel users to release packages with breaking changes alongside their dependents

cryptodev-2s:
This PR Compel users to release packages with breaking changes alongside their dependents.

Issue: [90](https://github.com/MetaMask/create-release-branch/issues/90)


src/release-specification.ts
```diff
@@ -301,6 +300,72 @@ export async function validateReleaseSpecification(
...
+            return false;
+          }
+
+          const dependentVersionSpecifierOrDirective =
+            unvalidatedReleaseSpecification.packages[
+              dependent.validatedManifest.name
+            ];
+
+          return (
+            dependentVersionSpecifierOrDirective !== SKIP_PACKAGE_DIRECTIVE &&
```
Gudahtt, line 332: In this case I think we do want to show this error if the user has `null` selected, because of the hazards.

src/release-specification.ts
```diff
@@ -301,6 +300,72 @@ export async function validateReleaseSpecification(
...
+        });
+
+        if (missingDependents.length > 0) {
+          errors.push({
+            message: [
+              `The following packages, which depends on released package ${packageName}, are missing.`,
+              missingDependents
+                .map((dependent) => `  - ${dependent.validatedManifest.name}`)
+                .join('\n'),
+              " Consider including them in the release spec so that they won't break in production.",
```
Gudahtt, line 350: Nit: Maybe better to describe this as a compatibility problem. It was a bit unclear from reading this _how_ they might break things, but "compatibility" is a way to get the point across without going into detail.
```suggestion
              ` Consider including them in the release spec so that they are compatible with the new ${packageName} version.`,
```

src/release-specification.ts
```diff
@@ -301,6 +300,72 @@ export async function validateReleaseSpecification(
...
+              dependentVersionSpecifierOrDirective,
+            ) &&
+            !isValidSemver(dependentVersionSpecifierOrDirective)
+          );
+        });
+
+        if (missingDependents.length > 0) {
+          errors.push({
+            message: [
+              `The following packages, which depends on released package ${packageName}, are missing.`,
```
Gudahtt, line 346: Nit: Superficial improvements to the message:
```suggestion
              `The following packages, which depend on released package '${packageName}', are missing.`,
```

src/release-specification.test.ts
Gudahtt, line 1: A few more test cases that would be nice to capture:

* A package dependent on another with a major version bump is set to `null` in the release spec (should throw)
* A package dependent on another with a major version bump is set to `intentionally-skip` in the release spec (should not throw)

Gudahtt (commented):
Looks great! Left a few suggestions, the most important of which is ensuring that packages set to `null` still produce this error

src/release-specification.test.ts
Gudahtt: Oh, we should also ensure these errors are not triggered for minor/patch bumps. Updating the descriptions for existing tests to make it clear that they only apply for major bumps would be good as well

src/release-specification.ts
```diff
@@ -301,6 +300,72 @@ export async function validateReleaseSpecification(
           });
         }
       }
+
+      // Check to compel users to release packages with breaking changes alongside their dependents
+      if (
+        versionSpecifierOrDirective === 'major' ||
+        (isValidSemver(versionSpecifierOrDirective) &&
+          diff(
+            pkg.validatedManifest.version,
```
Gudahtt, line 309: I noticed that the other PR checks that `pkg` is truthy first before using it. Would the same check be appropriate here? We are using it on this line

src/release-specification.test.ts
```diff
@@ -242,7 +242,13 @@ packages:
       await withSandbox(async (sandbox) => {
         const project = buildMockProject({
           workspacePackages: {
-            a: buildMockPackage('a'),
+            a: buildMockPackage('a', {
```
mcmire, line 245: Was this added as a sanity check? The `validateReleaseSpecification` function doesn't care what's inside of each package's manifest, it really just cares about the release spec, so this test doesn't need to be explicit about what's in dependencies. But I'm curious what your thought was here.

mcmire (commented):
Sorry it took me so long to get to this PR. I was unable to get to the end of this PR today, but would definitely like to take a closer look Monday.

src/release-specification.test.ts
cryptodev-2s: this was added just for sanity to show up everything still works even though we added the new check

cryptodev-2s: I just removed the change

src/release-specification.test.ts
```diff
@@ -683,5 +683,286 @@ ${releaseSpecificationPath}
...
+
+The release spec file has been retained for you to edit again and make the necessary fixes. Once you've done this, re-run this tool.
+
+${releaseSpecificationPath}
+`.trim(),
+        );
+      });
+    });
+
+    it("does not throw an error if packages in the release with a major version bump have their dependents via 'dependencies' with their version specified as 'intentionally-skip' in the release spec", async () => {
```
mcmire, line 853: This test and the previous test are great! Do we want to copy them and use `peerDependencies` instead of `dependencies`?

src/release-specification.test.ts
```diff
@@ -683,5 +683,286 @@ ${releaseSpecificationPath}
...
+        });
+        const releaseSpecificationPath = path.join(
+          sandbox.directoryPath,
+          'release-spec',
+        );
+        await fs.promises.writeFile(
+          releaseSpecificationPath,
+          YAML.stringify({
+            packages: {
+              a: 'major',
```
mcmire, line 767: It looks like in these tests, you're handling what happens when the version is bumped by using the word "major", but I think we will also want to test what happens when a literal version is used, e.g. if package `a` is 1.0.0 and the release spec contains "2.0.0" for `a`.

mcmire (commented):
This is good! A lot of the comments here have to do with simplified ways of doing the same thing or tweaks to wording, but there are some missing test cases as well.

src/release-specification.ts
```diff
@@ -301,6 +300,72 @@ export async function validateReleaseSpecification(
...
+            versionSpecifierOrDirective as string,
+          ) === 'major')
+      ) {
+        const missingDependents = Object.values(
+          project.workspacePackages,
+        ).filter((dependent) => {
+          const { dependencies, peerDependencies } =
```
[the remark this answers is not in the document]
cryptodev-2s: yes make sense I will wait for https://github.com/MetaMask/create-release-branch/pull/102 to be merged to take over the changes for replacement of unvalidatedManifest with validatedManifest

src/release-specification.test.ts
```diff
@@ -684,6 +684,700 @@ ${releaseSpecificationPath}
...
 
+    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "dependencies" are not listed in the release', async () => {
+      await withSandbox(async (sandbox) => {
+        const project = buildMockProject({
+          workspacePackages: {
+            a: buildMockPackage('a', {
+              hasChangesSinceLatestRelease: true,
+            }),
+            b: buildMockPackage('b', {
+              hasChangesSinceLatestRelease: false,
```
mcmire, line 695: See my note below about checking for changed dependents. In the context of this test, "b" doesn't have any changes, so I feel like this test shouldn't pass. What do you think about using `hasChangesSinceLatestRelease: true` in this and the other tests you've added? We may also want to add a new test to ensure that if "b" is NOT changed then the error doesn't get produced, but I can suggest that later in case you forget (there's a lot to keep track of here 😅 ).

src/release-specification.ts
```diff
@@ -317,6 +318,55 @@ export async function validateReleaseSpecification(
...
+              project.workspacePackages[possibleDependentName];
+            const { dependencies, peerDependencies } =
+              possibleDependent.validatedManifest;
+            return (
+              hasProperty(dependencies, packageName) ||
+              hasProperty(peerDependencies, packageName)
+            );
+          },
+        );
+        const missingDependentNames = dependentNames.filter((dependentName) => {
```
mcmire, line 340: Given the following scenario:

- There are two packages, A and B, where B depends on A
- A is listed with a major bump, but B is not listed

I think something might be missing: we're never checking whether there are any changes in B to release. 

Now... there's no reason why B _shouldn't_ have changes, because the version it's using for the A dependency should be bumped along with A. (If A is bumped from 1.0.0 to 2.0.0, then B's dependency on A should also be bumped to `~> 2.0.0`). However, I don't think we are ever enforcing that. That means it's possible for B to not have changes.

Should we add a new list here, similar to other lists we've created?

``` suggestion
        );
        const changedDependentNames = dependentNames.filter((dependentName) => {
          return project.workspacePackages[possibleDependentName]
            .hasChangesSinceLatestRelease;
        });
        const missingDependentNames = changedDependentNames.filter((dependentName) => {
```

Additionally, do you think it would be worth it to add some kind of check in a future PR which ensures that if workspace package A is being bumped to a major version X, and it is a major version, then all dependents / peer dependents of A are also bumped to `~^X.0.0`?

src/release-specification.ts
```diff
@@ -317,6 +318,55 @@ export async function validateReleaseSpecification(
...
+          return !unvalidatedReleaseSpecification.packages[dependentName];
+        });
+
+        if (missingDependentNames.length > 0) {
+          errors.push({
+            message: [
+              `The following packages, which depend on released package '${packageName}', are missing from the release spec.`,
```
mcmire, line 347: Wondering if we should clarify that it's the major bump that's causing this error:

``` suggestion
              `The following dependents of package '${packageName}', which is being released with a major version bump, are missing from the release spec.`,
```

src/release-specification.ts
```diff
@@ -210,23 +211,48 @@ export async function validateReleaseSpecification(
...
       const isInternalDependency = Object.values(
         project.workspacePackages,
-      ).some((pkg) => {
-        const { dependencies, peerDependencies } = pkg.validatedManifest;
+      ).some((workspacePackage) => {
+        const { dependencies, peerDependencies } =
+          workspacePackage.validatedManifest;
         return (
           hasProperty(dependencies, packageName) ||
           hasProperty(peerDependencies, packageName)
```
mcmire, line 232: To go along with the above, what about:

``` suggestion
          hasProperty(dependencies, missingChangedPackageName) ||
          hasProperty(peerDependencies, missingChangedPackageName)

File: src/release-specification.ts [hunk 13]
```diff
@@ -210,23 +211,48 @@ export async function validateReleaseSpecification(
...
       });
 
-      return (
-        (!hasProperty(unvalidatedReleaseSpecification.packages, packageName) ||
-          unvalidatedReleaseSpecification.packages[packageName] === null) &&
-        !isInternalDependency
-      );
+      const hasInternalDependencyWithBreakingChanges = Object.keys({
+        ...pkg.validatedManifest.dependencies,
+        ...pkg.validatedManifest.peerDependencies,
```
mcmire, line 238: To go along with the above, what about:

``` suggestion
        ...missingChangedPackage.validatedManifest.dependencies,
        ...missingChangedPackage.validatedManifest.peerDependencies,

File: src/release-specification.ts [hunk 14]
```diff
@@ -210,23 +211,48 @@ export async function validateReleaseSpecification(
...
+        .filter((dependency) => project.workspacePackages[dependency])
+        .some((dependency) => {
+          const internalDependencyVersionSpecifierOrDirective =
+            unvalidatedReleaseSpecification.packages[dependency];
+          return (
+            internalDependencyVersionSpecifierOrDirective &&
+            (internalDependencyVersionSpecifierOrDirective === 'major' ||
+              (isValidSemver(internalDependencyVersionSpecifierOrDirective) &&
+                diff(
+                  pkg.validatedManifest.version,
```
mcmire, line 249: To go along with the above, what about:

``` suggestion
                  missingChangedPackage.validatedManifest.version,

File: src/release-specification.ts [hunk 15]
```diff
@@ -317,6 +343,63 @@ export async function validateReleaseSpecification(
         }
       }
 
+      // Check to compel users to release packages with breaking changes alongside their dependents
+      if (
+        versionSpecifierOrDirective === 'major' ||
+        (isValidSemver(versionSpecifierOrDirective) &&
+          diff(pkg.validatedManifest.version, versionSpecifierOrDirective) ===
```
mcmire, line 350: `pkg` and `packageName` are also kind of confusing here, but there's nothing we can do about that right now. Perhaps we keep this in mind for a future refactor, though.

mcmire (approved):
I have a few more suggestions for making this code a bit easier to read, but I know I've also suggested refactoring this code, so considering you've been waiting on this PR for a while, that's something we can put off. The logic at least seems correct here, so feel free to merge!

src/release-specification.ts
cryptodev-2s: I did update pkg to changedPackage and packageName to changedPackageName

mcmire (approved):
Looks good! Again sorry for the long wait on this.

PR #102, opened 2023-09-28
Compel users to release new versions of dependencies alongside their dependents

cryptodev-2s:
This PR Compel users to release new versions of dependencies alongside their dependents

Issue: [91](https://github.com/MetaMask/create-release-branch/issues/91)


src/release-specification.ts
```diff
@@ -301,6 +299,58 @@ export async function validateReleaseSpecification(
...
+      ) {
+        const missingDependencies = Object.keys({
+          ...(pkg.unvalidatedManifest.dependencies || {}),
+          ...(pkg.unvalidatedManifest.peerDependencies || {}),
+        }).filter((dependency) => {
+          const dependencyVersionSpecifierOrDirective =
+            unvalidatedReleaseSpecification.packages[dependency];
+
+          return (
+            dependencyVersionSpecifierOrDirective !== SKIP_PACKAGE_DIRECTIVE &&
```
Gudahtt, line 318: We'll want to show this error if `null` is specified. The message suggests that only `intentionally-skip` should override the error.

src/release-specification.ts
```diff
@@ -301,6 +299,58 @@ export async function validateReleaseSpecification(
...
+          isValidSemver(versionSpecifierOrDirective))
+      ) {
+        const missingDependencies = Object.keys({
+          ...(pkg.unvalidatedManifest.dependencies || {}),
+          ...(pkg.unvalidatedManifest.peerDependencies || {}),
+        }).filter((dependency) => {
+          const dependencyVersionSpecifierOrDirective =
+            unvalidatedReleaseSpecification.packages[dependency];
+
+          return (
```
Gudahtt, line 317: Note that we should only show this warning if the dependency has unreleased changes. We'll need an additional filter step.

At the moment it seems to expect _all_ dependencies from being listed, even those not in the monorepo!

src/release-specification.ts
```diff
@@ -301,6 +299,58 @@ export async function validateReleaseSpecification(
...
+        });
+
+        if (missingDependencies.length > 0) {
+          errors.push({
+            message: [
+              `The following packages, which uses a released package ${packageName}, are missing.`,
+              missingDependencies
+                .map((dependency) => `  - ${dependency}`)
+                .join('\n'),
+              " Consider including them in the release spec so that they won't break in production.",
```
Gudahtt, line 336: Nit: I've attempted to re-word this to explain the risk more explicitly:
```suggestion
              ` These packages may have changes that '${packageName}' relies upon. Consider including them in the release spec.`,
```

src/release-specification.ts
```diff
@@ -301,6 +299,58 @@ export async function validateReleaseSpecification(
...
+              `  If you are ABSOLUTELY SURE that this won't occur, however, and want to postpone the release of a package, then list it with a directive of "intentionally-skip". For example:`,
```
Gudahtt, line 337: Nit: I've attempted to re-word this to make the risks more explicit:
```suggestion
              ` These packages may have changes that '${packageName}' relies upon. Consider including them in the release spec.`,
              `  If you are ABSOLUTELY SURE that these packages are safe to omit, however, and want to postpone the release of a package, then list it with a directive of "intentionally-skip". For example:`,
```

src/release-specification.ts
```diff
@@ -301,6 +299,58 @@ export async function validateReleaseSpecification(
...
+          return (
+            dependencyVersionSpecifierOrDirective !== SKIP_PACKAGE_DIRECTIVE &&
+            dependencyVersionSpecifierOrDirective !==
+              INTENTIONALLY_SKIP_PACKAGE_DIRECTIVE &&
+            !hasProperty(
+              IncrementableVersionParts,
+              dependencyVersionSpecifierOrDirective,
+            ) &&
+            !isValidSemver(dependencyVersionSpecifierOrDirective)
+          );
```
mcmire, line 326: Following Mark's comment, I'm wondering whether we can make this conditional simpler than it is, knowing that:

- if the release spec is missing a package, then attempting to look it up should return `undefined`.
- if a version was never specified for a package and it's not being intentionally skipped, then it should be `null`.

Therefore we should be able to say:

``` suggestion
          return (
            dependencyVersionSpecifierOrDirective === undefined ||
            dependencyVersionSpecifierOrDirective === SKIP_PACKAGE_DIRECTIVE
          );
```

src/release-specification.ts
mcmire: Right, that's a good point. We can use `project.workspacePackages` to look up the package object for a `dependency`, and once we have it we can access its `.hasChangesSinceLatestRelease` property to know whether it's changed.

src/release-specification.ts
mcmire: I think that this change makes sense. We'll probably want to be consistent about whether we use "depends on" or "relies on", but this is fine for now — I can give this PR another pass.

src/release-specification.test.ts
```diff
@@ -678,6 +678,71 @@ Your release spec could not be processed due to the following issues:
...
+          }),
+        );
+
+        await expect(
+          validateReleaseSpecification(project, releaseSpecificationPath),
+        ).rejects.toThrow(
+          `
+Your release spec could not be processed due to the following issues:
+
+* The following packages, which have changed since their latest release, are missing.
```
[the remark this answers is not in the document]
cryptodev-2s: @mcmire I agree, and I made the change

src/release-specification.ts
cryptodev-2s: actually the error should be thrown in case dependencyVersionSpecifierOrDirective === undefined or dependencyVersionSpecifierOrDirective === null I updated the PR to !dependencyVersionSpecifierOrDirective in this case

src/release-specification.ts
```diff
@@ -301,6 +313,55 @@ export async function validateReleaseSpecification(
...
+          if (
+            !project.workspacePackages[dependency]?.hasChangesSinceLatestRelease
+          ) {
+            return false;
+          }
+
+          const dependencyVersionSpecifierOrDirective =
+            unvalidatedReleaseSpecification.packages[dependency];
+
+          return !dependencyVersionSpecifierOrDirective;
```
mcmire, line 337: Since this returns a boolean, is this just as readable?

``` suggestion
          project.workspacePackages[dependency]?.hasChangesSinceLatestRelease &&
            !unvalidatedReleaseSpecification.packages[dependency]
```

src/release-specification.ts
```diff
@@ -211,9 +211,23 @@ export async function validateReleaseSpecification(
...
+          return (
+            (dependencies && hasProperty(dependencies, packageName)) ||
+            (peerDependencies && hasProperty(peerDependencies, packageName))
+          );
+        },
+      );
+
+      return (
+        (!hasProperty(unvalidatedReleaseSpecification.packages, packageName) ||
+          unvalidatedReleaseSpecification.packages[packageName] === null) &&
```
mcmire, line 227: Since we're using the unvalidated release specification — and to be consistent with how you are making this check below — can we say this?

``` suggestion
        !unvalidatedReleaseSpecification.packages[packageName] &&
```

src/release-specification.test.ts
```diff
@@ -683,5 +683,160 @@ ${releaseSpecificationPath}
         );
       });
     });
+
+    it("throws if there are any packages not listed in the release which have changed, and are being defined as 'dependencies' by other packages which are listed", async () => {
```
mcmire, line 687: Nit: This sentence might read better if you remove the comma:

``` suggestion
    it("throws if there are any packages not listed in the release which have changed and are being defined as 'dependencies' by other packages which are listed", async () => {
```

src/release-specification.test.ts
```diff
@@ -683,5 +683,160 @@ ${releaseSpecificationPath}
...
+
+The release spec file has been retained for you to edit again and make the necessary fixes. Once you've done this, re-run this tool.
+
+${releaseSpecificationPath}
+`.trim(),
+        );
+      });
+    });
+
+    it("throws if there are any packages not listed in the release which have changed, and are being defined as 'peerDependencies' by other packages which are listed", async () => {
```
mcmire, line 742: Nit: This sentence might read better if you remove the comma:

``` suggestion
    it("throws if there are any packages not listed in the release which have changed and are being defined as 'peerDependencies' by other packages which are listed", async () => {
```

src/release-specification.ts
```diff
@@ -301,6 +313,55 @@ export async function validateReleaseSpecification(
...
+      // Check to compel users to release new versions of dependencies alongside their dependents
+      if (
+        pkg &&
+        versionSpecifierOrDirective &&
+        (hasProperty(IncrementableVersionParts, versionSpecifierOrDirective) ||
+          isValidSemver(versionSpecifierOrDirective))
+      ) {
+        const missingDependencies = Object.keys({
+          ...(pkg.unvalidatedManifest.dependencies || {}),
+          ...(pkg.unvalidatedManifest.peerDependencies || {}),
```
mcmire, line 326: Note that if you incorporate `dependencies` and `peerDependencies` into the `validatedManifest` then you can default these properties to an empty object, so then you could simplify this to:

``` suggestion
          ...pkg.validatedManifest.dependencies,
          ...pkg.validatedManifest.peerDependencies,

File: src/release-specification.ts [hunk 12]
```diff
@@ -301,6 +313,55 @@ export async function validateReleaseSpecification(
...
+          const dependencyVersionSpecifierOrDirective =
+            unvalidatedReleaseSpecification.packages[dependency];
+
+          return !dependencyVersionSpecifierOrDirective;
+        });
+
+        if (missingDependencies.length > 0) {
+          errors.push({
+            message: [
+              `The following packages, which are dependencies of the package '${packageName}' being released, are missing from the release spec.`,
```
mcmire, line 343: Since they may also be peer dependencies, what do you think about:

``` suggestion
              `The following packages, which are dependencies or peer dependencies of the package '${packageName}' being released, are missing from the release spec.`,
```

src/release-specification.test.ts
```diff
@@ -683,5 +683,160 @@ ${releaseSpecificationPath}
...
+
+The release spec file has been retained for you to edit again and make the necessary fixes. Once you've done this, re-run this tool.
+
+${releaseSpecificationPath}
+`.trim(),
+        );
+      });
+    });
+
+    it("does not throw when a package defined as 'dependencies' by a listed package in the release has not changed", async () => {
```
mcmire, line 797: Nice! Should we mention that the package that hasn't changed is also not listed?

``` suggestion
    it("does not throw when any packages defined as 'dependencies' by a listed package in the release are not listed but have not changed", async () => {
```

mcmire: Also, what do you think about a copy of this test but for `peerDependencies` instead of `dependencies`?

mcmire (commented):
This is looking better! Just some user experience tweaks and attempts to simplify the logic so it's a bit easier to parse.

tests/unit/helpers.ts
mcmire, line 1: Note: The changes here may not be necessary following the comments I made above about `validatedManifest`, since you can specify `validatedManifest` in tests instead of `unvalidatedManifest`.

src/package-manifest.ts
```diff
@@ -256,6 +266,63 @@ export function readPackageManifestPrivateField(
...
+    if (!isTruthyString(pkgName)) {
+      return false;
+    }
+
+    if (isValidPackageManifestVersionField(version)) {
+      return false;
+    }
+  }
+
+  return true;
```
mcmire, line 289: This function returns a boolean already, so would this be as easy to read?

``` suggestion
  return Object.entries(depsValue).every(([pkgName, version]) => {
    return isTruthyString(pkgName) && isValidPackageManifestVersionField(version);
  });

File: src/package-manifest.ts [hunk 15]
```diff
@@ -29,6 +29,8 @@ export type ValidatedPackageManifest = {
   readonly [PackageManifestFieldNames.Version]: SemVer;
   readonly [PackageManifestFieldNames.Private]: boolean;
   readonly [PackageManifestFieldNames.Workspaces]: string[];
+  readonly [PackageManifestFieldNames.Dependencies]: Record<string, string>;
+  readonly [PackageManifestFieldNames.PeerDependencies]: Record<string, string>;
```
mcmire, line 33: Ah, this would make sense — but it seems like these can be found under different names:

``` suggestion
  readonly [PackageManifestDependenciesFieldNames.Production]: Record<string, string>;
  readonly [PackageManifestDependenciesFieldNames.Peer]: Record<string, string>;

File: src/package-manifest.ts [hunk 16]
```diff
@@ -83,6 +85,14 @@ const schemata = {
     validate: isValidPackageManifestPrivateField,
     errorMessage: 'must be true or false (if present)',
   },
+  [PackageManifestFieldNames.Dependencies]: {
```
mcmire, line 88:
``` suggestion
  [PackageManifestDependenciesFieldNames.Production]: {

File: src/package-manifest.ts [hunk 17]
```diff
@@ -83,6 +85,14 @@ const schemata = {
...
+    validate: isValidPackageManifestDependenciesField,
+    errorMessage: 'must be a valid dependencies field',
+  },
+  [PackageManifestFieldNames.PeerDependencies]: {
```
mcmire, line 92:
``` suggestion
  [PackageManifestDependenciesFieldNames.Peer]: {

File: src/package-manifest.ts [hunk 18]
```diff
@@ -256,6 +266,63 @@ export function readPackageManifestPrivateField(
...
+ * @param fieldName - The field name "dependencies" or "peerDependencies".
+ * @returns The value of the "dependencies" or "peerDependencies" field.
+ * @throws If the value of the field is not valid.
+ */
+export function readPackageManifestDependenciesField(
+  manifest: UnvalidatedPackageManifest,
+  parentDirectory: string,
+  fieldName:
+    | PackageManifestFieldNames.Dependencies
+    | PackageManifestFieldNames.PeerDependencies,
```
mcmire, line 307:
``` suggestion
    | PackageManifestDependenciesFieldNames.Production
    | PackageManifestDependenciesFieldNames.Peer,

File: src/package-manifest.ts [hunk 19]
```diff
@@ -281,12 +348,24 @@ export async function readPackageManifest(manifestPath: string): Promise<{
     unvalidated,
     parentDirectory,
   );
+  const dependencies = readPackageManifestDependenciesField(
+    unvalidated,
+    parentDirectory,
+    PackageManifestFieldNames.Dependencies,
```
mcmire, line 354:
``` suggestion
    PackageManifestDependenciesFieldNames.Production,

File: src/package-manifest.ts [hunk 20]
```diff
@@ -281,12 +348,24 @@ export async function readPackageManifest(manifestPath: string): Promise<{
...
+  );
+  const peerDependencies = readPackageManifestDependenciesField(
+    unvalidated,
+    parentDirectory,
+    PackageManifestFieldNames.PeerDependencies,
```
mcmire, line 359:
``` suggestion
    PackageManifestDependenciesFieldNames.Peer,

File: src/package-manifest.ts [hunk 21]
```diff
@@ -281,12 +348,24 @@ export async function readPackageManifest(manifestPath: string): Promise<{
...
+  );
 
   const validated = {
     [PackageManifestFieldNames.Name]: name,
     [PackageManifestFieldNames.Version]: version,
     [PackageManifestFieldNames.Workspaces]: workspaces,
     [PackageManifestFieldNames.Private]: privateValue,
+    [PackageManifestFieldNames.Dependencies]: dependencies,
+    [PackageManifestFieldNames.PeerDependencies]: peerDependencies,
```
mcmire, line 368:
``` suggestion
    [PackageManifestDependenciesFieldNames.Production]: dependencies,
    [PackageManifestDependenciesFieldNames.Peer]: peerDependencies,

**Review by mcmire — COMMENTED** [review]
Suggestions to fix the TypeScript errors.

File: src/package-manifest.ts [hunk 22]
```diff
@@ -256,6 +269,57 @@ export function readPackageManifestPrivateField(
...
 
+/**
+ * Type guard to ensure that the value of the "dependencies" or "peerDependencies" field of a manifest is
+ * valid.
+ *
+ * @param depsValue - The value to check.
+ * @returns Whether the value is has valid values.
+ */
+function isValidPackageManifestDependenciesField(
+  depsValue: Record<string, string> = {},
```
mcmire, line 280: The `= {}` implies that `isValidPackageManifestDependenciesField` can be called without any arguments, like `isValidPackageManifestDependenciesField()`. Would that happen? I'm wondering if it makes more sense to remove this. Also, we don't technically know what type of thing is being passed yet — we are verifying that it's a `Record<string, string>`, but it could be anything. So we can communicate that by using `unknown`:

``` suggestion 
  depsValue: unknown,
```

src/release-specification.ts
```diff
@@ -210,28 +210,45 @@ export async function validateReleaseSpecification(
...
-      unvalidatedReleaseSpecification.packages[packageName] === null,
+  const packageNamesToReportMissing = changedPackageNames.filter(
+    (packageName) => {
+      const isInternalDependency = Object.values(
+        project.workspacePackages,
+      ).some((pkg) => {
+        const { dependencies, peerDependencies } = pkg.validatedManifest;
+        return (
+          (dependencies && hasProperty(dependencies, packageName)) ||
+          (peerDependencies && hasProperty(peerDependencies, packageName))
```
mcmire, line 221: I think you can drop the `dependencies &&` and `peerDependencies &&` checks now since we know now they will never be `null` or `undefined`:

``` suggestion
          hasProperty(dependencies, packageName) ||
          hasProperty(peerDependencies, packageName)

mcmire (commented):
This looks great! Just two more things and then I think we can call this good:

mcmire (approved):
Sorry this took so long and thanks for sticking with it. Looks good now!

Issue #88, opened 2023-08-15
Reorder workflow to update changelogs first


mcmire:
Currently, the workflow of this tool is cumbersome. As soon as the tool is run, it will essentially ask the user to choose the next version of the packages changed. However, to do this, the user must first review the changes that have taken place to those packages. This can be quite painful to do, and it repeats work that this tool does already.

Instead, this tool should add all detected changes to the appropriate changelogs first, filing them under the "Unreleased" header. The user can then review these changes, following the referenced pull requests as necessary, and assign an appropriate version to each package.

If we rearranged the workflow in this manner, we would have to be careful not to introduce any new pain points. In this case, we would want to make sure that the user can assign a version to a package easily. This is not just a matter of updating the `package.json`, but also consists of renaming the "Unreleased" header in the changelog to match the new version and updating any URLs at the bottom of the file to point to the future Git tag. If the user has to do this for multiple packages, this will end up being cumbersome. Perhaps a subcommand can be added to the `auto-changelog` tool to support doing this more easily.

mcmire:
One thing to keep in mind with this modified workflow is that it could introduce another pain point. When the tool applied the release spec, it would expect all of the changes it previously made to the changelogs to be intact, and it would also expect package versions to be the same as when it was first run. That is, if the user made further changes to changelogs or versions before applying the release spec, this would confuse the tool, and it may not be able to successfully proceed. However, perhaps we wouldn't need to completely account for this. All the tool would need to worry about is whatever entries happen to be in Unreleased changelog sections and whatever the current version of packages happen to be. If any changelogs or versions have been changed in such a way as to be unparseable, then the tool will break (as it does right now).

mcmire:
Related to the above comment, however, if the user bumps a version of a package prior to generating the release spec, they are effectively indicating that they want to release that package, and so when the release spec is generated, then we could populate the release spec accordingly (i.e. if the user bumps a package version from 1.0.1 to 1.1.0 then we detect it as a minor change and then populate the release spec with `minor` for that package). The user would be free to change this; if they do, then the tool would use the original version of the package as a base, where this is determined by looking at the base branch of the current branch.

xinnanyemm:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Shared-Libraries-621e46b4d7103800171d1b02/issues/MetaMask/create-release-branch/88?planning-poker) with Zenhub @cryptodev-2s @Gudahtt @mikesposito

[...]

PR #109, opened 2023-11-16
Reorder workflow to update changelogs first

cryptodev-2s:
Reorder workflow to update changelogs first

Issue: [88](https://github.com/MetaMask/create-release-branch/issues/88)

### DIFF
--- a/src/monorepo-workflow-operations.ts
+++ b/src/monorepo-workflow-operations.ts
@@ -10,12 +10,13 @@ import { determineEditor } from './editor';
 import { ReleaseType } from './initial-parameters';
 import { Project } from './project';
 import { planRelease, executeReleasePlan } from './release-plan';
-import { captureChangesInReleaseBranch } from './repo';
+import { createReleaseBranch, captureChangesInReleaseBranch } from './repo';
 import {
   generateReleaseSpecificationTemplateForMonorepo,
   waitForUserToEditReleaseSpecification,
   validateReleaseSpecification,
 } from './release-specification';
+import { updateChangedPackagesChangelog } from './package';
 
 /**
  * For a monorepo, the process works like this:
@@ -64,6 +65,15 @@ export async function followMonorepoWorkflow({
   stdout: Pick<WriteStream, 'write'>;
   stderr: Pick<WriteStream, 'write'>;
 }) {
+  const { version: newReleaseVersion, firstRun } = await createReleaseBranch({
+    project,
+    releaseType,
+  });
+
+  if (firstRun) {
+    await updateChangedPackagesChangelog({ project, stderr });
+  }
+
   const releaseSpecificationPath = path.join(
     tempDirectoryPath,
     'RELEASE_SPEC.yml',
@@ -115,11 +125,11 @@ export async function followMonorepoWorkflow({
   const releasePlan = await planRelease({
     project,
     releaseSpecification,
-    releaseType,
+    newReleaseVersion,
   });
   await executeReleasePlan(project, releasePlan, stderr);
   await removeFile(releaseSpecificationPath);
   await captureChangesInReleaseBranch(project.directoryPath, {
-    releaseVersion: releasePlan.newVersion,
+    releaseVersion: newReleaseVersion,
   });
 }
--- a/src/package.ts
+++ b/src/package.ts
@@ -236,6 +236,66 @@ export async function readMonorepoWorkspacePackage({
  *
  * @param args - The arguments.
  * @param args.project - The project.
+ * @param args.stderr - A stream that can be used to write to standard error.
+ * @returns The result of writing to the changelog.
+ */
+export async function updateChangedPackagesChangelog({
+  project: { repositoryUrl, workspacePackages },
+  stderr,
+}: {
+  project: Pick<
+    Project,
+    'directoryPath' | 'repositoryUrl' | 'workspacePackages'
+  >;
+  stderr: Pick<WriteStream, 'write'>;
+}): Promise<void> {
+  await Promise.all(
+    Object.values(workspacePackages)
+      .filter(
+        ({ hasChangesSinceLatestRelease }) => hasChangesSinceLatestRelease,
+      )
+      .map(async (pkg) => {
+        let changelogContent;
+
+        try {
+          changelogContent = await readFile(pkg.changelogPath);
+        } catch (error) {
+          if (isErrorWithCode(error) && error.code === 'ENOENT') {
+            stderr.write(
+              `${pkg.validatedManifest.name} does not seem to have a changelog. Skipping.\n`,
+            );
+            return;
+          }
+
+          throw error;
+        }
+
+        const newChangelogContent = await updateChangelog({
+          changelogContent,
+          isReleaseCandidate: false,
+          projectRootDirectory: pkg.directoryPath,
+          repoUrl: repositoryUrl,
+          tagPrefixes: [`${pkg.validatedManifest.name}@`, 'v'],
+        });
+
+        if (newChangelogContent) {
+          await writeFile(pkg.changelogPath, newChangelogContent);
+        } else {
+          stderr.write(
+            `Changelog for ${pkg.validatedManifest.name} was not updated as there were no updates to make.`,
+          );
+        }
+      }),
+  );
+}
+
+/**
+ * Updates the changelog file of the given package using
+ * `@metamask/auto-changelog`. Assumes that the changelog file is located at the
+ * package root directory and named "CHANGELOG.md".
+ *
+ * @param args - The arguments.
+ * @param args.project - The project.
  * @param args.packageReleasePlan - The release plan for a particular package in
  * the project.
  * @param args.stderr - A stream that can be used to write to standard error.
--- a/src/release-plan.ts
+++ b/src/release-plan.ts
@@ -1,6 +1,5 @@
 import type { WriteStream } from 'fs';
 import { SemVer } from 'semver';
-import { ReleaseType } from './initial-parameters';
 import { debug } from './misc-utils';
 import { Package, updatePackage } from './package';
 import { Project } from './project';
@@ -48,33 +47,25 @@ export type PackageReleasePlan = {
 
 /**
  * Uses the release specification to calculate the final versions of all of the
- * packages that we want to update, as well as a new release name.
+ * packages that we want to update.
  *
  * @param args - The arguments.
  * @param args.project - Information about the whole project (e.g., names of
  * packages and where they can found).
  * @param args.releaseSpecification - A parsed version of the release spec
  * entered by the user.
- * @param args.releaseType - The type of release ("ordinary" or "backport"),
- * which affects how the version is bumped.
+ * @param args.newReleaseVersion - The new release version.
  * @returns A promise for information about the new release.
  */
 export async function planRelease({
   project,
   releaseSpecification,
-  releaseType,
+  newReleaseVersion,
 }: {
   project: Project;
   releaseSpecification: ReleaseSpecification;
-  releaseType: ReleaseType;
+  newReleaseVersion: string;
 }): Promise<ReleasePlan> {
-  const newReleaseVersion =
-    releaseType === 'backport'
-      ? `${project.releaseVersion.ordinaryNumber}.${
-          project.releaseVersion.backportNumber + 1
-        }.0`
-      : `${project.releaseVersion.ordinaryNumber + 1}.0.0`;
-
   const rootReleasePlan: PackageReleasePlan = {
     package: project.rootPackage,
     newVersion: newReleaseVersion,
--- a/src/repo.ts
+++ b/src/repo.ts
@@ -1,9 +1,12 @@
 import path from 'path';
 import {
+  debug,
   runCommand,
   getStdoutFromCommand,
   getLinesFromCommand,
 } from './misc-utils';
+import { ReleaseType } from './initial-parameters';
+import { Project } from './project';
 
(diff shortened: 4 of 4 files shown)


src/monorepo-workflow-operations.ts
```diff
@@ -64,6 +65,15 @@ export async function followMonorepoWorkflow({
   stdout: Pick<WriteStream, 'write'>;
   stderr: Pick<WriteStream, 'write'>;
 }) {
+  const { version: newReleaseVersion, firstRun } = await createReleaseBranch({
+    project,
+    releaseType,
+  });
+
+  if (firstRun) {
```
mcmire, line 73: Ah, good to only do this on the first run. That makes sense.

src/repo.ts
```diff
@@ -186,17 +184,85 @@ export async function captureChangesInReleaseBranch(
...
+    project.directoryPath,
+    'rev-parse',
+    ['--abbrev-ref', 'HEAD'],
+  );
+
+  if (currentBranchName === releaseBranchName) {
+    debug(`Already on ${releaseBranchName} branch.`);
+    return {
+      version: newReleaseVersion,
+      firstRun: false,
```
mcmire, line 233: The purpose of this file was that it would provide some general, single-purpose helper functions for working with a repo. However, the existence of this property makes this more of a higher level function, something that does multiple things rather than one thing. What do you think about moving it to `monorepo-workflow-operations.ts`?

src/monorepo-workflow-operations.ts
```diff
@@ -64,6 +65,15 @@ export async function followMonorepoWorkflow({
   stdout: Pick<WriteStream, 'write'>;
   stderr: Pick<WriteStream, 'write'>;
 }) {
+  const { version: newReleaseVersion, firstRun } = await createReleaseBranch({
```
mcmire, line 68: What do you think about updating the tests for this file to make sure that these steps are performed?

src/release-plan.ts
mcmire, line 1: I see that in this file, `planRelease` returns an object where `shouldUpdateChangelog` is `true`. Do we need this property at all anymore?

src/package.ts
mcmire, line 1: I see that in this file, `updatePackage` still calls `updatePackageChangelog`. Now that changelogs are generated up front, does that need to happen anymore?

mcmire (commented):
Good changes! But there are some tests missing, and I think there's some code we can drop as well.

src/package.test.ts
```diff
@@ -465,27 +467,58 @@ describe('package', () => {
...
+          - Add \`isNewFunction\` ([#2](https://repo.url/compare/package/pull/2))
+
+          ## [1.0.0] - 2020-01-01
+          ### Changed
+          - Something else
+
+          [Unreleased]: https://repo.url/compare/package@2.0.0...HEAD
+          [1.0.0]: https://repo.url/releases/tag/package@1.0.0
+        `,
+        );
```
mcmire, line 492: The indentation here seems a bit misleading. What about:

``` suggestion
          outdent`
            # Changelog
            All notable changes to this project will be documented in this file.
            The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
            and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
            ## [Unreleased]
            ### Uncategorized
            - Add \`isNewFunction\` ([#2](https://repo.url/compare/package/pull/2))
            ## [1.0.0] - 2020-01-01
            ### Changed
            - Something else
            [Unreleased]: https://repo.url/compare/package@2.0.0...HEAD
            [1.0.0]: https://repo.url/releases/tag/package@1.0.0
          `),
```

src/package.test.ts
```diff
@@ -465,27 +467,58 @@ describe('package', () => {
...
+        - Add \`isNewFunction\` ([#2](https://repo.url/compare/package/pull/2))
+
+        ## [1.0.0] - 2020-01-01
+        ### Changed
+        - Something else
+
+        [Unreleased]: https://repo.url/compare/package@2.0.0...HEAD
+        [2.0.0]: https://repo.url/compare/package@1.0.0...package@2.0.0
+        [1.0.0]: https://repo.url/releases/tag/package@1.0.0
+      `);
```
mcmire, line 521: Similar as above. Perhaps we want:

``` suggestion
        expect(newChangelogContent).toBe(outdent`
          # Changelog
          All notable changes to this project will be documented in this file.
          The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
          and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
          ## [Unreleased]
          ## [2.0.0]
          ### Uncategorized
          - Add \`isNewFunction\` ([#2](https://repo.url/compare/package/pull/2))
          ## [1.0.0] - 2020-01-01
          ### Changed
          - Something else
          [Unreleased]: https://repo.url/compare/package@2.0.0...HEAD
          [2.0.0]: https://repo.url/compare/package@1.0.0...package@2.0.0
          [1.0.0]: https://repo.url/releases/tag/package@1.0.0
        `);

File: src/package.ts [hunk 6]
```diff
@@ -304,18 +354,17 @@ export async function updatePackage({
...
 
   await writeJsonFile(pkg.manifestPath, {
     ...pkg.unvalidatedManifest,
     version: newVersion,
   });
 
-  if (shouldUpdateChangelog) {
-    await updatePackageChangelog({ project, packageReleasePlan, stderr });
-  }
+  await migrateChangelogUnreleasedChangesToRelease({
```
mcmire, line 364: Replacing this makes the description for this function a bit outdated. Right now it reads:

```
 * Updates the package as per the instructions in the given release plan by
 * replacing the `version` field in the manifest and adding a new section to the
 * changelog for the new version of the package.
```

It seems like we might want something like:

```
 * Updates the package by replacing the `version` field in the manifest
 * according to the one in the given release plan. Also updates the
 * changelog by migrating changes in the Unreleased section to the section
 * representing the new version.
```

src/package.test.ts
```diff
@@ -15,12 +16,14 @@ import {
...
   readMonorepoWorkspacePackage,
   updatePackage,
+  updatePackageChangelog,
 } from './package';
 import * as fsModule from './fs';
 import * as packageManifestModule from './package-manifest';
 import * as repoModule from './repo';
 
-jest.mock('@metamask/auto-changelog');
+const outdent = _outdent({ trimTrailingNewline: false });
```
mcmire, line 25: Nice that a package can do this for us. Heads up that we already have a version of this in `tests/functional/helpers/utils.ts`, called `normalizeMultilineString` — we might want to replace that with `outdent` at some point.

src/package.test.ts
```diff
@@ -451,7 +453,7 @@ describe('package', () => {
       });
     });
 
-    it('updates the changelog of the package if requested to do so and if the package has one', async () => {
+    it('should migrate all unreleased changes to a release section.', async () => {
```
mcmire, line 456: Nit: We don't typically use "should" in tests (see our [guideline about naming tests](https://github.com/MetaMask/contributor-docs/blob/main/docs/unit-testing.md#use-it-to-specify-the-desired-behavior-for-the-code-under-test)):

``` suggestion
    it('migrates all unreleased changes to a release section', async () => {

File: src/project.test.ts [hunk 9]
```diff
@@ -120,4 +130,147 @@ describe('project', () => {
       });
     });
   });
+  describe('resetUnreleasedPackagesChangelog', () => {
```
mcmire, line 133: This goes along with a previous comment, but what do you think about `restoreChangelogsForSkippedPackages`? If so, then it would be good to have this name reflect that:

``` suggestion

  describe('restoreChangelogsForSkippedPackages', () => {
```

or at least reflect the current name:

``` suggestion

  describe('restoreUnreleasedPackagesChangelog', () => {
```

src/project.test.ts
```diff
@@ -120,4 +130,147 @@ describe('project', () => {
...
+    it('should reset changelog for packages with changes not included in release', async () => {
+      const project = buildMockProject({
+        rootPackage: buildMockPackage('monorepo'),
+        workspacePackages: {
+          a: buildMockPackage('a', {
+            hasChangesSinceLatestRelease: true,
+          }),
+          b: buildMockPackage('b', {
+            hasChangesSinceLatestRelease: false,
```
mcmire, line 142: If you wanted to test that packages that don't have changes don't get restored, it would make sense to set `false` here, but that doesn't seem to be what this test is about, so this could be potentially misleading. Thoughts on setting this to `true` and testing that "b" gets reset along with "c"?

src/project.test.ts
```diff
@@ -120,4 +130,147 @@ describe('project', () => {
...
+            a: IncrementableVersionParts.minor,
+          },
+          path: '/path/to/release/specs',
+        },
+      });
+
+      expect(restoreFilesSpy).not.toHaveBeenCalledWith(
+        '/path/to/packages/b',
+        'checkout',
+        ['--', 'CHANGELOG.md'],
```
mcmire, line 202: Hmm, should this be:

``` suggestion
        project.directoryPath,
        'main',
        ['/path/to/packages/b/CHANGELOG.md'],
```

src/project.test.ts
```diff
@@ -120,4 +130,147 @@ describe('project', () => {
...
+          path: '/path/to/release/specs',
+        },
+      });
+
+      expect(restoreFilesSpy).toHaveBeenCalledWith(
+        project.directoryPath,
+        'main',
+        ['/path/to/packages/c/CHANGELOG.md'],
+      );
+    });
```
mcmire, line 168: Should we also add a test for when a package is listed, but as "intentionally-skip"?

src/project.test.ts
```diff
@@ -120,4 +130,147 @@ describe('project', () => {
...
+          path: '/path/to/release/specs',
+        },
+      });
+
+      expect(restoreFilesSpy).not.toHaveBeenCalledWith(
+        '/path/to/packages/b',
+        'checkout',
+        ['--', 'CHANGELOG.md'],
+      );
+    });
```
mcmire, line 204: Should we also add a test for when the changelog file does not exist?

mcmire (commented):
Nice work. I think all of the pieces are in place, it's just a matter of polishing them a bit. I made some suggestions below.

src/project.ts
```diff
@@ -125,3 +128,83 @@ export async function readProject(
...
+    return changelogPaths;
+  }, []);
+
+  if (unreleasedPackagesChangelogPaths.length > 0) {
+    await restoreFiles(
+      directoryPath,
+      defaultBranch,
+      unreleasedPackagesChangelogPaths,
+    );
+  }
```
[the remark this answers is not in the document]
cryptodev-2s: @mcmire 

```
            !releaseSpecification.packages[name] ||
            releaseSpecification.packages[name] === INTENTIONALLY_SKIP_PACKAGE_DIRECTIVE
```
is not relevant here as restore occurs after releaseSpec validation so releaseSpecification.packages includes only package to release

src/project.test.ts
cryptodev-2s: @mcmire same as comment above intentionally-skip is not relevant here as this the final release specification which includes only packages to release

tests/functional/helpers/remote-repo.ts
```diff
@@ -14,7 +14,7 @@ export default class RemoteRepo extends Repo {
     await fs.promises.mkdir(this.getWorkingDirectoryPath(), {
       recursive: true,
     });
-    await this.runCommand('git', ['init', '--bare']);
+    await this.runCommand('git', ['init', '-b', 'main', '--bare']);
```
cryptodev-2s, line 17: @mcmire functional tests where failing due to default branch was set to master on pipeline although passes in Local that's why I did this change

mcmire: Makes sense!

src/functional.test.ts
```diff
@@ -639,8 +653,30 @@ describe('create-release-branch (functional)', () => {
           workspaces: {
             '.': ['packages/*'],
           },
+          createInitialCommit: false,
```
mcmire, line 656: Now that we confirm that changelogs exist before attempting to restore them, do we need these changes any longer?

src/command-line-arguments.ts
```diff
@@ -44,6 +45,12 @@ export async function readCommandLineArguments(
       type: 'boolean',
       default: false,
     })
+    .option('default-branch', {
```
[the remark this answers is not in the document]
cryptodev-2s: yes make sense as this is a one time config it more relevant to be on the release.config.json I don't how this work but happy to do it in a another PR

src/repo.ts
```diff
@@ -170,13 +173,8 @@ export async function getRepositoryHttpsUrl(
 }
 
 /**
- * This function does three things:
- *
- * 1. Stages all of the changes which have been made to the repo thus far and
+ * This function stages all of the changes which have been made to the repo thus far and
```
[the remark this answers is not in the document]
cryptodev-2s: I just did saw this comment but I already  did rename to  ```commitAllChanges``` let me know if it's okay ?

src/package.test.ts
cryptodev-2s: test fail on this case as parseChangelog does add these empty lines

mikesposito (dismissed):
Looks good! I only have a non-blocking suggestion

mcmire (dismissed):
I just tested this out on the `core` repo, and it seems everything works as designed, so well done. There are more workflow improvements we could possibly make, but we can add those in future PRs.

PR #112, opened 2023-11-30
Quote `editor.path` argument

rekmarks:
The `editor.path` string can contain spaces, e.g. `/Applications/Visual Studio Code.app/...`. This string is passed to `execa()`, which attempts to execute it as a command in the user's shell. On some systems, this string must be quoted, or else the shell will interpret the substring until the first space as the command, and either error or execute the wrong thing.

### DIFF
--- a/src/release-specification.ts
+++ b/src/release-specification.ts
@@ -123,7 +123,7 @@ export async function waitForUserToEditReleaseSpecification(
   );
 
   const promiseForEditorCommand = runCommand(
-    editor.path,
+    `"${editor.path}"`,
     [...editor.args, releaseSpecificationPath],
     {
       stdio: 'inherit',


rekmarks:
@legobeat hmm, that's interesting. The weirdness here is that this has not been a problem for anyone who has used this tool, except myself as described above. Specifically, the error I got was:

```text
/bin/sh: /Applications/Visual: No such file or directory
ErrorWithCause: Encountered an error while waiting for the release spec to be edited.
    at wrapError (.../node_modules/@metamask/create-release-branch/dist/misc-utils.js:86:23)
    at waitForUserToEditReleaseSpecification (.../node_modules/@metamask/create-release-branch/dist/release-specification.js:93:42)
...
```

I already tried my change locally and it fixes the error. However, according to [this source](https://www.grymoire.com/Unix/Quote.html), we should probably single-quote paths:

>If you want to work with files with spaces or special characters in the filename, you may have to use quotes. For instance, if you wanted to create a file with a space in the name, you could use the following:
>`% cp /dev/null 'a file with spaces in the name'`

I'm discussing with @mcmire where we're going to introduce quoting / escaping. I'm leaving this in draft until it's ready for another round of reviews.

rekmarks:
@legobeat, single-quoting seems to work fine on my system:

`/bin/zsh`:
```
'echo' 'foo'
foo
```

`/bin/sh`:
```
'echo' 'foo'
foo
```

And this would be consistent with [this source](https://mywiki.wooledge.org/Quotes):
> Single quotes: '...' removes the special meaning of every character between the quotes. Everything inside single quotes becomes a literal string. The only character that you can't safely enclose in single quotes is a single quote.

rekmarks:
Closing in favor of #121.

PR #115, opened 2023-12-03
Disable git commit GPG signing in tests

rekmarks:
The tests on `main` fail on my machine due to some issue with GPG signing, which I have globally enabled by default. This PR disables GPG signing for git commits in tests to avoid this issue.


mcmire (commented):
Makes sense!

mcmire (approved):

PR #117, opened 2023-12-04
Replace outdent with our own implementation

mcmire:
Currently we use the `outdent` package so that tests that involve writing multi-line strings to files look a bit nicer. This package doesn't play nice in an ESM context, and we actually already have equivalents in `normalizeWhitespaceString` and `buildChangelog` that we can use instead. So this commit removes `outdent` in favor of these helpers.

### DIFF
--- a/tests/helpers.ts
+++ b/tests/helpers.ts
@@ -60,7 +60,7 @@ export async function withSandbox(fn: (sandbox: Sandbox) => any) {
   try {
     await fn({ directoryPath });
   } finally {
-    await rimraf(directoryPath);
+    await fs.promises.rm(directoryPath, { force: true, recursive: true });
   }
 }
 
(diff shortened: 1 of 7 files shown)


tests/helpers.ts
```diff
@@ -60,7 +59,7 @@ export async function withSandbox(fn: (sandbox: Sandbox) => any) {
   try {
     await fn({ directoryPath });
   } finally {
-    await rimraf(directoryPath);
+    await fs.promises.rm(directoryPath, { force: true, recursive: true });
```
rekmarks, line 62: Why switch from `rimraf`?

rekmarks: Nothing in particular, just curious!

rekmarks (approved):
LGTM!

PR #119, opened 2023-12-05
Don't recommend initial use of intentionally-skip

mcmire:
Currently we say in the releas spec that if the user wants to skip a package, they can specify "intentionally-skip" in place of the version. We actually don't want first-time users to know about this directive up front, because it suppresses errors that would ordinarily be produced if you omit a package from the release that you shouldn't. Rather, we want to only suggest this in the error message.

Hence, this commit removes the reference to "intentionally-skip" from the release spec. It also adds some more instructions for first-time users.

### DIFF
--- a/src/release-specification.test.ts
+++ b/src/release-specification.test.ts
@@ -45,23 +45,38 @@ describe('release-specification', () => {
 
       expect(template).toStrictEqual(
         `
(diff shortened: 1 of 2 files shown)


cryptodev-2s (approved):
LGTM!

PR #121, opened 2023-12-06
Update execa

rekmarks:
This PR updates `execa` to `^8.0.1`. Since `execa@>=6.0.0` is ESM-only and `jest` only has experimental ESM support (https://github.com/jestjs/jest/pull/10976), this required switching from `ts-jest` to `babel-jest`. To minimize dependency transpilation, the ESM packages that are necessary to transpile are enumerated in `jest.config.js`.

This version of `execa` includes [automatic escaping of shell arguments](https://github.com/sindresorhus/execa/tree/v8.0.1#execafile-arguments-options), which was the entire point of #112, #113, and this PR.

The state of ESM support in the Node.js ecosystem is absolutely horrible, and I would not recommend further migrations for the time being. We should continue to dual-release our packages and avoid ESM-only dependencies until the ecosystem has matured. For details see the above `jest` issue and https://github.com/nodejs/node/issues/37648.


rekmarks:
@SocketSecurity ignore get-stream@8.0.1

The new author appears to be a chosen collaborator of `sindresorhus`, who still owns the repo and the package.

PR #122, opened 2023-12-06
Move auto-changelog to deps, pin to ~3.3.0

mcmire:
We started to use `@metamask/auto-changelog` at runtime in a recent commit. As it turns out, recent versions of this tool contain breaking changes that we didn't document, and so we need to rely on an older version. Adding this to `dependencies` ensures that we are doing so, even if the project where this tool is run against a project that also also uses `@metamask/auto-changelog` (but a different version).

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@metamask/action-utils": "^1.0.0",
+    "@metamask/auto-changelog": "~3.3.0",
     "@metamask/utils": "^8.2.1",
     "debug": "^4.3.4",
     "execa": "^5.1.1",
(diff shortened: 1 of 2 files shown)


mcmire:
For more context, this PR in `auto-changelog` seems to have changed the behavior of `updateChangelog` so that it does nothing if `isReleaseCandidate` is `false`: https://github.com/MetaMask/auto-changelog/pull/158. That conflicts with the expectations this tool makes.

desi (approved):

mcmire:
@MajorLift That would be great — whenever you have time. I'm really curious why this didn't break any of the tests. I suspect we have some missing test cases. In the meantime I've filed a ticket here: https://github.com/MetaMask/create-release-branch/issues/124

MajorLift:
> I'm really curious why this didn't break any of the tests. 

Oh that's easy. I don't believe we have any tests for `update-changelog`! I'll make a PR for the fix asap and then look into adding some unit tests as well.

PR #106, opened 2023-11-08
Update Package Reader to Accept Range Versions in Dependencies

cryptodev-2s:
The recent merges from https://github.com/MetaMask/create-release-branch/pull/102 and https://github.com/MetaMask/create-release-branch/pull/101 have introduced a bug related to package version validation. Specifically, the issue arises when handling range versions, which are currently not permitted by the validation logic.

PR #125, opened 2023-12-15
Support 'workspace:^' notation in dependency versions

cryptodev-2s:
### Overview

This pull request addresses the issue where the 'workspace:^' notation was not being recognized as a valid specifier for dependency/peerDependencies versions in our package.json files.

### Changes Made

- Modified `isValidPackageManifestVersionField` to recognize `workspace:^` as a valid version.

### DIFF
--- a/src/package-manifest.ts
+++ b/src/package-manifest.ts
@@ -139,7 +139,7 @@ export function readPackageManifestNameField(
 function isValidPackageManifestVersionField(
   version: unknown,
 ): version is string {
-  return isTruthyString(version);
+  return isTruthyString(version) && semver.validRange(version) !== null;
 }
 
 /**
@@ -156,8 +156,7 @@ function isValidPackageManifestDependencyValue(
   version: unknown,
 ): version is string {
   return (
-    isValidPackageManifestVersionField(version) &&
-    (semver.validRange(version) !== null || version === 'workspace:^')
+    isValidPackageManifestVersionField(version) || version === 'workspace:^'
   );
 }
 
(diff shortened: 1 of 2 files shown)


src/package-manifest.ts
```diff
@@ -139,7 +139,10 @@ export function readPackageManifestNameField(
 function isValidPackageManifestVersionField(
```
mcmire, line 139: I see that this function is called `isValidPackageManifestVersionField`, but it seems that we are using this for not only checking that the "version" field of a `package.json` is correct, but also that the version for a dependency is correct. A package's version can't be `workspace:^`, so perhaps we want to leave this function unchanged and then copy it to a new function called `isValidPackageManifestDependencyValue` where we can make this change? Then we can have `isValidPackageManifestDependenciesField` use this new function instead.

cryptodev-2s: Yes indeed didn't pay attention to I removed the semver valid range check for the initial function as it was not there it was added later in this commit https://github.com/MetaMask/create-release-branch/pull/106

mcmire: Oh that's okay! I think it makes a lot of sense to keep the `isValidSemver` check in `isValidPackageManifestVersionField`. No reason to change that. It's just that the `workspace:^` check doesn't apply there, it only applies to dependency versions.

mcmire (approved):
I tested this out on `snaps` and this bug goes away. There is another bug that prevents `create-release-branch` from running, but that's because this tool doesn't support private packages yet. That's a separate issue and I'll make a ticket for it. Anyway, this looks good on its own though!

mcmire:
Follow up to this — it seems there were some empty directories in `packages/` and that caused the tool to throw an error. I've logged a ticket about this here: https://github.com/MetaMask/create-release-branch/issues/128. But it doesn't seem like this deters the Snaps team from using this tool, it's just a minor inconvenience.

PR #145, opened 2024-05-17
Fix constraints and update yarn lock at the end of release process

cryptodev-2s:
## Description

In our current release process, we often encounter issues with dependency constraints, outdated yarn.lock files, and duplicated dependencies. These issues can lead to inconsistencies in our release process. Additionally, these steps are frequently needed whenever we release our packages, making the process cumbersome and error-prone.

## Solution

This PR adds a new step to our release tool to streamline and automate the process. This step will:

1. Fix Constraints: Ensure that all package versions adhere to the specified constraints,-.
2. Update the Yarn Lockfile: Refresh the yarn.lock file to include the latest dependency resolutions..
3. Deduplicate Dependencies: Remove duplicate dependencies to streamline the dependency tree, which helps in reducing the overall package size and avoiding multiple versions of the same dependency.

Implementation Details

- Constraint Fixing: Introduces a command to enforce and fix version constraints for dependencies.
- Yarn Lockfile Update: Runs yarn install to regenerate the yarn.lock file with the latest dependency resolutions.
- Dependency Deduplication: Utilizes the yarn deduplicate command to clean up duplicated dependencies in the lockfile.
- Release Tool Update: Integrates these steps into our release tool to automate the process, making it more fluid and less error-prone.


cryptodev-2s:
@kanthesha yes for the test failing in the testing configuration I need to add a constraints file, currently working on it.

For the intentionally skip, yes the ticket doesn't update changelog there's another ticket for updating changelog for bumbed packages. The scope of this ticket is to fix constraints and update lock only.

tests/functional/helpers/local-monorepo.ts
```diff
@@ -159,7 +159,27 @@ export default class LocalMonorepo<
   protected async afterCreate() {
     await super.afterCreate();
 
-    await this.writeJsonFile('package.json', { private: true });
+    await this.writeJsonFile('package.json', {
+      private: true,
+      packageManager: 'yarn@3.2.1',
+    });
+
+    const constraintsProContent = `
```
mcmire, line 167: Hmm... this constraints file doesn't really check anything I don't think? I believe that Yarn looks for certain predicates to exist. Perhaps this just file just needs to be present?

mcmire (commented):
Is there a functional test we can add to confirm that the workflow bug we are trying to solve is fixed?

mcmire:
Also can we update the PR description to describe the problem that we are solving here and how the changes solve that problem instead of just describing the changes? That would help us understand the context behind this PR in the future.

tests/functional/helpers/local-monorepo.ts
cryptodev-2s: It checks that each package has a name and a version field
it's required for the constraints fix otherwise it throws `This package doesn't seem to be present in your lockfile; run "yarn install" to update the lockfile`

cryptodev-2s:
> Also can we update the PR description to describe the problem that we are solving here and how the changes solve that problem instead of just describing the changes? That would help us understand the context behind this PR in the future.

I updated the description

src/yarn-commands.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @throws An execa error object if the command fails in some way.
+ */
+export async function fixConstraints(
+  repositoryDirectoryPath: string,
+  stderr: Pick<WriteStream, 'write'>,
+): Promise<void> {
+  const version = await getYarnVersion();
+  const majorVersion = parseInt(version.split('.')[0], 10);
+
+  if (majorVersion >= 2) {
```
mcmire, line 29: I think we can safely assume that the Yarn version is >= 2. We don't use this tool for any other projects at the moment. So perhaps we can drop this check? (The `--no-immutable` option to `yarn` and the `yarn dedupe` commands already assume Yarn >= 2 anyway.)

tests/functional/helpers/local-monorepo.ts
mcmire: That's a strange error when the constraints file doesn't exist. I would have expected something like "constraints.pro not found".

I just tried removing these lines and running the tests and they don't seem to fail with the error you describe 🤔

mcmire:
Just repeating my comment above about adding a functional test here. I think the problem that running `yarn` was intended to solve was so that the release PR wouldn't fail CI. CI runs `yarn --immutable` and then it runs `yarn constraints` as a part of the lint step, so it seems like we ought to have a test that runs the tool, then runs `yarn --immutable`, then `yarn constraints` and expects both of these commands to return an exit code of 0. This way we can be sure that we've solved the original problem. Is this possible to add?

tests/functional/helpers/local-monorepo.ts
cryptodev-2s: we can remove the constraints file only the plugin installation is required, but don't we want to test that constraints check really works as this constraints at least checks that each package has a version and name, we can add more check ?

src/functional.test.ts
```diff
@@ -640,6 +642,263 @@ describe('create-release-branch (functional)', () => {
       );
     });
 
+    it('updates the dependency version in package "b" when package "a" version is bumped', async () => {
```
mcmire, line 645: `create-release-branch` runs `yarn constraints --fix` and doesn't care what the constraints are exactly. So it isn't necessary the case that `create-release-branch` always updates dependencies like this — it really depends on what the project's constraints file does (a project may have different constraints, in theory). That said, I'm okay with using this test name since it points to why we decided to add `yarn constraints --fix` in the first place. If we ever want to run constraints for a different reason we may have to adjust this test name to be more accurate, but I don't anticipate that happening any time soon.

mcmire (approved):
Sorry for the long delay in reviewing this. Looks good to me.

PR #165, opened 2025-01-24
Fix repo links in updated changelogs to match `auto-changelog`

### LINKED ISSUE #72: Changelogs are updated with incorrect URLs if developer uses a non-standard URL for `origin` remote
This tool uses the URL of the repository to populate links in the changelog when updating it. The logic to obtain this URL, however, is different from the logic that the `auto-changelog` executable uses when it validates the changelog. Namely, [this tool uses `git config --get remote.origin.url`](https://github.com/MetaMask/create-release-branch/blob/0eccf5cf05ad269a2aa47ed925e9b920c4f68b72/src/repo.ts#L141-L151), whereas [`auto-changelog` reads it from the `package.json`](https://github.com/MetaMask/auto-changelog/blob/839be83cf62b364a8094690d6a992a335037d8be/src/repo.ts#L19-L44) (in the case of a Yarn v3 project, anyway). We should make these the same so that `create-release-branch` does not update the changelog in such a way that makes it invalid.

mcmire:
When this tool populates the Unreleased section of the changelog, it does not just rewrite that section, but rather, it rewrites the entire changelog. That behavior comes from the `updateChangelog` function in `@metamask/auto-changelog`. This function takes the changes to apply, but it also takes a repository URL, which it will use to construct links for all of the releases in the changelog.

Unfortunately, the repository URL that this tool passes to `updateChangelog` is not always the same URL that `auto-changelog` uses under the hood. Specifically, if the developer used a different URL to clone the repo than is listed in `package.json` under `repository.url`, then when creating a new release branch, all of the links in the changelog will be modified to reflect this new URL. This is incorrect, as the resulting changelog will not pass validation when run through `auto-changelog validate`, forcing the developer to revert the changes to the links in order to merge the release PR.

This commit addresses this pain point by updating the logic used to obtain the repository URL to better match what is in `@metamask/auto-changelog`. This means that the aforementioned links in the changelog should stay the same when creating a new release branch.

Fixes #72.


src/repo.ts
```diff
@@ -123,52 +125,6 @@ async function getFilesChangedSince(
...
- * - git@github.com:OrganizationName/RepositoryName.git
- *
- * If the URL of the "origin" remote matches neither pattern, an error is
- * thrown.
- *
- * @param repositoryDirectoryPath - The path to the repository directory.
- * @returns The HTTPS URL of the repository, e.g.
- * `https://github.com/OrganizationName/RepositoryName`.
- */
-export async function getRepositoryHttpsUrl(
```
mcmire, line 141: I split up this function into two pieces so that it could be tested better: I extracted the code to get the repository URL to `project.ts` and the string manipulation logic to `misc-utils.ts`.

src/project.ts
```diff
@@ -132,6 +143,62 @@ export async function readProject(
...
+ * @returns The HTTPS URL of the repository, e.g.
+ * `https://github.com/OrganizationName/RepositoryName`.
+ * @throws If the URL obtained via `package.json` or the "origin" remote is in
+ * in invalid format.
+ */
+export async function getValidRepositoryUrl(
+  packageManifest: UnvalidatedPackageManifest,
+  repositoryDirectoryPath: string,
+): Promise<string> {
+  // Set automatically by NPM or Yarn 1.x
```
mcmire, line 176: Although handling NPM or Yarn 1.x projects is not strictly necessary here, because `create-release-branch` only supports monorepos, and all of our monorepos use Yarn Modern, I figured I'd include it here because I want to copy this function back to `auto-changelog` after this PR is merged.

For context the original source for this logic is here: https://github.com/MetaMask/auto-changelog/blob/ef3e86e15b0de7061856a53fd18c4f38e898f5e8/src/repo.ts#L20

src/project.ts
```diff
@@ -132,6 +143,62 @@ export async function readProject(
...
+  }
+
+  if (
+    isPlainObject(packageManifest.repository) &&
+    typeof packageManifest.repository.url === 'string'
+  ) {
+    return convertToHttpsGitHubRepositoryUrl(packageManifest.repository.url);
+  }
+
+  const gitConfigUrl = await getStdoutFromCommand(
```
mcmire, line 194: Given that most of our projects ought to have a `repository` field in `package.json`, I am not sure that this is really needed, but I figured I'd leave it here.

I tracked down the original source for this logic, and it comes from here: https://github.com/MetaMask/action-monorepo-release-pr/commit/af8d7504aa7b8cd4a65a81625839adf3afc3cbb9#diff-b2d6c9ae0590887f3d213032d057be61bd3e628650fb7ba99391bdc84587e454R25

src/misc-utils.ts
```diff
@@ -172,3 +172,38 @@ export async function getLinesFromCommand(
...
+ * - https://github.com/OrganizationName/RepositoryName
+ * - git@github.com:OrganizationName/RepositoryName.git
+ *
+ * If the URL does not match either of these patterns, an error is thrown.
+ *
+ * @param url - The URL to convert.
+ * @returns The HTTPS URL of the repository, e.g.
+ * `https://github.com/OrganizationName/RepositoryName`.
+ */
+export function convertToHttpsGitHubRepositoryUrl(url: string): string {
```
mcmire, line 190: This logic is not in `auto-changelog`, but is in this repo, so we're actually doing extra. I was curious why there was a distinction, so I did a bit of digging:

- In building `create-release-branch`, I took a lot of code from `action-create-release-pr`
- We used to have a separate repo called `action-monorepo-release-pr`, and that got merged into `action-create-release-pr`
- In particular this logic comes from here: https://github.com/MetaMask/action-monorepo-release-pr/commit/af8d7504aa7b8cd4a65a81625839adf3afc3cbb9#diff-b2d6c9ae0590887f3d213032d057be61bd3e628650fb7ba99391bdc84587e454R25

mcmire: Also, I've simplified the original logic to take a full-on regex approach. It seemed that the code wanted to go there.

src/functional.test.ts
```diff
@@ -77,13 +77,9 @@ describe('create-release-branch (functional)', () => {
             },
           });
 
-          expect(await environment.readJsonFile('package.json')).toStrictEqual({
```
mcmire, line 80: These tests are updated because there's now an extra `repository` field in `package.json`. That isn't really relevant to these tests, so I've simplified the assertions to only consider `name` and `version`.

src/misc-utils.ts
mcmire: I will probably end up copying this back to `action-create-release-pr` so that it is consistent between the two.

cryptodev-2s (approved):
LGTM!

Issue #163, opened 2024-11-27
Create web UI


mcmire:
The release process is arduous, and the CLI does not do a good job of helping users navigate through it. A lot of developers are more comfortable working with GUIs than command-line tools. A web UI could help dramatically improve the contributor experience with regard to creating releases for the `core` repo.

## Acceptance Criteria

- When the developer runs `create-release-branch`, a new tab opens in the user's browser that displays a web UI. The UI presents a form that allows the developer to select which packages they want to release and how to bump those packages (much like the CLI today).
- If the developer deselects a package that ought to be included (because it has a peer dependency that was bumped, or because it is a dependency of another package) then the developer is informed immediately and is disallowed to submit the form unless they confirm they've checked there would be no impact (perhaps by selecting `intentionally skip` as an option, for simplicity, unless we can think of a better pattern).
- When the developer submits the form, the changes are applied and the developer is put on the new release branch. For each package selected, the version of the package is bumped accordingly and commit messages that have occurred since the previous release are copied to the changelog for the package (just like what happens now).

PR #166, opened 2025-02-20
feat: release UI beta

cryptodev-2s:
# The Release UI Beta

You can test the new feature in `create-release-branch` locally with **Core** by following these steps:

1️⃣ Checkout the Feature Branch
```sh
git checkout cryptodev2s/release-ui-beta
```

3️⃣ Run the Command in `create-release-branch`

```sh
yarn && yarn build
```

4️⃣ Update Core’s package.json

> Modify package.json in Core to point create-release-branch to the local build.

4️⃣ Run the Command in Core
```sh
yarn && yarn create-release-branch -i
```

Fixes: https://github.com/MetaMask/create-release-branch/issues/163


cryptodev-2s:
@SocketSecurity ignore-all

@SocketSecurity ignore npm/methods@1.1.2
@SocketSecurity ignore npm/raw-body@2.5.2
@SocketSecurity ignore npm/encodeurl@2.0.0
@SocketSecurity ignore npm/body-parser@1.20.3
@SocketSecurity ignore npm/send@0.19.0
@SocketSecurity ignore npm/express@4.21.2

src/ui/App.tsx
```diff
@@ -0,0 +1,596 @@
...
+
+type ReleaseType =
+  | 'major'
+  | 'minor'
+  | 'patch'
+  | 'intentionally-skip'
+  | 'custom'
+  | string;
+
+function App() {
```
Mrtenz, line 20: I'd try breaking this component up, since there is a lot going on, and it's a bit hard to follow right now.

cryptodev-2s: yes currently doing this

src/ui/PackageItem.tsx
```diff
@@ -0,0 +1,157 @@
...
+        packageDependencyErrors[pkg.name] &&
+        packageDependencyErrors[pkg.name].missingDependencies.length > 0
+          ? 'border-red-500'
+          : ''
+      }`}
+    >
+      <div className="flex items-start gap-3">
+        <div className="pt-1">
+          <input
+            type="checkbox"
```
mcmire, line 62: I know that it was suggested to add a checkbox for each package item, but now it could be potentially confusing: does checking the checkbox add the package to the release or does it just select it so that a bulk action can be applied? 

I wonder if it makes sense to hide the checkboxes by default. Then we always show the bulk action bar, but if nothing is checked we show a button that says "Select multiple packages".

```
+------------------------------
| |Select multiple packages|
+------------------------------
```

You click that, it shows checkboxes next to each package item, and the bulk action bar switches to:

```
+--------------------------------------------------------------
| Bulk action for 3 packages: |Major| |Minor| |Patch| |Skip|
+--------------------------------------------------------------
```

Once you've bulk applied an action then the checkboxes will be hidden and the bulk action bar will go back to:

```
+------------------------------
| |Select multiple packages|
+------------------------------
```

src/ui/App.tsx
```diff
@@ -0,0 +1,380 @@
...
+      alert('Failed to submit selections. Please try again.');
+    } finally {
+      setIsSubmitting(false);
+    }
+  };
+
+  const fetchChangelog = async (packageName: string): Promise<void> => {
+    setLoadingChangelogs((prev) => ({ ...prev, [packageName]: true }));
+    try {
+      const response = await fetch(`/api/changelog?package=${packageName}`);
```
mcmire, line 242: I notice that this returns the entire changelog for the package. It'd be nice if in a future we could capture and return the unreleased changes instead. I feel like those might be more useful than providing the entire changelog, because it's the unreleased changes that dictate how you want to bump a package, or whether you need to include it in the release.

src/ui/App.tsx
```diff
@@ -0,0 +1,380 @@
...
+      disabled={isDisabled}
+      className={`mt-6 px-4 py-2 rounded ${
+        isDisabled
+          ? 'bg-blue-300 cursor-not-allowed'
+          : 'bg-blue-500 hover:bg-blue-600'
+      } text-white`}
+    >
+      {isSubmitting
+        ? 'Processing Release (This may take a few minutes)...'
+        : 'Submit Release Selections'}
```
mcmire, line 42: Maybe we can be a bit more straightforward about what will happen?

``` suggestion
        ? 'Creating Release Branch (This may take a few minutes)...'
        : 'Create Release Branch'}

File: src/ui/PackageItem.tsx [hunk 5]
```diff
@@ -0,0 +1,157 @@
...
+  isSelected,
+  onToggleSelect,
+}: PackageItemProps) {
+  return (
+    <div
+      key={pkg.name}
+      id={`package-${pkg.name}`}
+      className={`border p-4 rounded-lg ${
+        selections[pkg.name] && selections[pkg.name] !== 'intentionally-skip'
+          ? 'border-gray-500'
```
mcmire, line 50: Maybe we add `border-2` in addition to or instead of a darkened border so it's a little more clear what's selected?

src/ui/App.tsx
```diff
@@ -0,0 +1,380 @@
...
+  }
+
+  if (isSuccess) {
+    return (
+      <div className="fixed inset-0 bg-white flex items-center justify-center">
+        <div className="text-center p-8">
+          <h2 className="text-2xl font-bold mb-4">Release Successful!</h2>
+          <p className="mb-6">You can now close this window.</p>
+          <button
+            onClick={() => window.close()}
```
mcmire, line 300: This `window.close` doesn't seem to work. I get an error message in devtools: `Scripts may close only the windows that were opened by them.`

Maybe we don't need this button?

src/dirname.ts
```diff
@@ -0,0 +1,14 @@
+import { fileURLToPath } from 'url';
+import { dirname } from 'path';
+
+const __filename = fileURLToPath(import.meta.url);
```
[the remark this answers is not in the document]
cryptodev-2s: Actually I extracted the file because otherwise jest fails to run the main test check the second commit for reference

src/interactive-ui.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  app.get('*', (_req, res) => {
+    res.sendFile(join(UI_BUILD_DIR, 'index.html'));
+  });
+
+  const server = app.listen(port, () => {
+    stdout.write(`Interactive UI server running at http://localhost:${port}\n`);
+  });
+
+  return new Promise((_, reject) => {
```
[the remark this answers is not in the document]
cryptodev-2s: indeed I did forget to resolve the promise, it has been fixed now

src/ui/App.tsx
cryptodev-2s: Yes I agree with this

src/ui/App.tsx
```diff
@@ -0,0 +1,380 @@
...
+  const handleSelectionChange = (
+    packageName: string,
+    value: ReleaseType | '',
+  ): void => {
+    setSelections((prev) => {
+      if (value === '') {
+        const { [packageName]: _, ...rest } = prev;
+        const { [packageName]: __, ...remainingErrors } =
+          packageDependencyErrors;
+        setPackageDependencyErrors(remainingErrors);
```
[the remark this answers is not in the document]
cryptodev-2s: yes I don't how I missed this

src/ui/App.tsx
```diff
@@ -0,0 +1,380 @@
...
+    }
+  };
+
+  const handleSelectionChange = (
+    packageName: string,
+    value: ReleaseType | '',
+  ): void => {
+    setSelections((prev) => {
+      if (value === '') {
+        const { [packageName]: _, ...rest } = prev;
```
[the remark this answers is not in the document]
cryptodev-2s: It omits a package from the selections, removing it from the specs after it has been added if we no longer intend to release it or mark it as skipped.

src/ui/index.html
```diff
@@ -0,0 +1,32 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Create Release Branch Interactive UI</title>
+    <link
+      rel="icon"
+      href="https://metamask.github.io/test-dapp/metamask-fox.svg"
+    />
+    <script type="importmap">
+      {
+        "imports": {
+          "react": "https://esm.sh/react@19",
```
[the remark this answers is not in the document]
cryptodev-2s: Yes, exactly. I plan to remove this behavior and generate a build that includes only the required chunks.

src/ui.ts
```diff
@@ -0,0 +1,305 @@
...
+      } catch (error) {
+        stderr.write(`Release error: ${error}\n`);
+        res.status(400).send('Invalid request');
+      }
+    },
+  );
+
+  app.get('*', (_req, res) => {
+    res.sendFile(join(UI_BUILD_DIR, 'index.html'));
+  });
```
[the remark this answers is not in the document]
mcmire: It is probably okay to ignore this, this is just going to be used in development.

src/ui/index.html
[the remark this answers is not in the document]
cryptodev-2s: this has been removed we now build the UI with vite check this [commit](https://github.com/MetaMask/create-release-branch/pull/166/commits/edbfa01ed5c0ba745293e5190dfa2c191941b220)

src/ui/PackageItem.tsx
```diff
@@ -0,0 +1,161 @@
...
+        </div>
+      </div>
+
+      {packageDependencyErrors[pkg.name] && (
+        <div className="mt-2 p-3 bg-red-50 border border-red-200 rounded">
+          <div className="flex-grow">
+            {packageDependencyErrors[pkg.name].missingDependencies.length >
+              0 && (
+              <DependencyErrorSection
+                title="Missing Dependencies"
```
mcmire, line 121: I wonder if we should include some extra instructions here.

For missing dependencies / peer dependencies, can we say something like this:

> The following packages are dependencies or peer dependencies of \<package name\>. Because they may have introduced new changes that \<package name\> is now using, you need to verify whether to include them in the release. 
>
> To do this, look at the change history for each package and compare it with the change history for \<package name\>. If \<package name\> uses any new changes from a package, then you need to include it by bumping its version. If you have confirmed that the changes to a package do not affect \<package name\>, you may omit it from the release by choosing "Skip" instead.

For missing dependents, can we say something like:

> Because \<package name\> is being released with a new major version, to prevent peer dependency warnings in consuming projects, all of the following packages which list <package name> as a peer dependency need to be included in the release. Please choose new versions for these packages. If for some reason you feel it is safe to omit a package you may choose "Skip".

If you feel like this is too long we could also add some Markdown documentation to this repo and link to that? What do you think?

src/ui/App.tsx
```diff
@@ -0,0 +1,383 @@
...
+          </h2>
+          <p className="mb-6">You can now close this window.</p>
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <div className="container mx-auto px-4 py-8">
+      <h1 className="text-3xl font-bold mb-6">Create New Core Release</h1>
```
mcmire, line 305: I wonder if we should include some instructions at the top. Maybe a few paragraphs like this?

> **How to use this tool**
>
> Listed below are the packages in this monorepo which have unreleased changes. You do not need to release all packages; you may release a subset. To add a new package, you first need to decide how to bump its version. Look over the changelog or commit history to understand the nature of the changes that have been made to the package. Then you can select a version from the dropdown next to the package. If there are breaking changes, select "Major"; if there are new features, select "Minor"; and if there are only non-functional backward-compatible changes, select "Patch".
>
> Some packages are linked via dependencies, and because of this you may be asked to include other packages in the release you were not anticipating. We've done our best to explain why, but feel free to reach out to the Wallet Framework team if you have any questions.

src/release-specification.ts
```diff
@@ -167,72 +168,162 @@ export async function waitForUserToEditReleaseSpecification(
...
+export function findMissingUnreleasedDependents(
+  project: Project,
+  packageName: string,
+  unvalidatedReleaseSpecificationPackages: Record<string, string | null>,
+): string[] {
+  const dependentNames = Object.keys(project.workspacePackages).filter(
+    (possibleDependentName) => {
+      const possibleDependent =
+        project.workspacePackages[possibleDependentName];
+      const { dependencies, peerDependencies } =
```
mcmire, line 187: I am just now realizing that this checks not just peer dependencies but also dependencies. In the [original ticket](https://github.com/MetaMask/create-release-branch/issues/90) I did say that we ought to include dependencies, but I don't think this is actually that useful. If we have a package A that depends on another package B and B has breaking changes, I think it's okay to just release B without releasing A. If a client has both A and B as direct dependencies, and they just upgrade B, then yes, we may have multiple versions of A in the dependency tree, but that's not a big deal. This is really only a big deal if A has a _peer dependency_ on B, because then if the client upgrades B they may get a peer dependency warning.

So... I think we should drop `dependencies` here. Maybe we can do this in a new PR before we merge this, or maybe we can do it after. The instructions I suggested we add if there are missing dependents do _not_ mention anything around regular (non-peer) dependencies, so I don't want to forget about it.

What do you think we should do?

cryptodev-2s: Yes make sense. I applied the change in this PR if that works for you!

src/ui/style.css
```diff
@@ -0,0 +1,5 @@
+@import 'tailwindcss';
+
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
```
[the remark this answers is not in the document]
cryptodev-2s: yes good catch

src/release-specification.ts
cryptodev-2s: Yes I can do both before/after. let me prepare a separate PR to fix this.

cryptodev-2s: Done in https://github.com/MetaMask/create-release-branch/pull/170

mcmire (approved):
Looks good to me! 🎉🎉

PR #167, opened 2025-02-26
Migrate to Vitest

Mrtenz:
Work in progress.

This migrates from Jest to Vitest, since the latter has much better compatibility with ESM.


cryptodev-2s:
@mcmire is this PR now ready?

Issue #164, opened 2025-01-13
If a monorepo package's peer dependencies are bumped and the package is not being released, it should be flagged in the validation step


mcmire:
Multiple times I've reviewed a release PR in the `core` repo where a package gets bumped by a major, this causes peer dependencies on that package in other packages to get bumped, but the tool does not advise that those dependent packages also get bumped. We should investigate whether this case is being handled properly in the validation code.

mcmire:
It seems like what we need to do to solve this issue is to change how we perform the current validations. We need to simulate the changes to the package manifests that applying the release spec would make, and the perform the validation on the simulated state of the monorepo rather than on the current state of the monorepo.

This may be essentially what we are doing already in the release UI, and if so, we should probably wait until we have that UI ready, and then add a new validation. Otherwise, we will have to spend some time creating that simulated state.

My guess is that this ticket is between a 5 and an 8.

PR #173, opened 2025-03-11
fix: include unchanged dependents in release validation

cryptodev-2s:
## Description
This PR fixes an important visibility issue in our release process where packages that should be considered for release due to peer dependency updates weren't being surfaced to users during the release workflow if they had no changes since their latest release.

### Technical Details
- Modified `findMissingUnreleasedDependents` to consider all dependent packages, not just those with changes since their latest release
- This ensures that all dependencies are properly validated during the release process, regardless of their change status

Fixes https://github.com/MetaMask/create-release-branch/issues/164

### DIFF
--- a/src/release-specification.test.ts
+++ b/src/release-specification.test.ts
@@ -657,7 +657,7 @@ ${releaseSpecificationPath}
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "peerDependencies" have no changes since their latest release and are not listed in the release', async () => {
+    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "peerDependencies" are not listed in the release, even if they have no changes', async () => {
       await withSandbox(async (sandbox) => {
         const project = buildMockProject({
           workspacePackages: {
@@ -767,7 +767,7 @@ ${releaseSpecificationPath}
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using a literal version, but any of their dependents defined as "peerDependencies" have no changes since their latest release and are not listed in the release', async () => {
+    it('throws if there are any packages in the release with a major version bump using a literal version, but any of their dependents defined as "peerDependencies" are not listed in the release, even if they have no changes', async () => {
       await withSandbox(async (sandbox) => {
         const project = buildMockProject({
           workspacePackages: {
@@ -878,7 +878,7 @@ ${releaseSpecificationPath}
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using the word "major", but their dependents via "peerDependencies" have no changes since their latest release and have their version specified as null in the release spec', async () => {
+    it('throws if there are any packages in the release with a major version bump using the word "major", but their dependents via "peerDependencies" have their version specified as null in the release spec, even if they have no changes', async () => {
       await withSandbox(async (sandbox) => {
         const project = buildMockProject({
           workspacePackages: {
@@ -990,7 +990,7 @@ ${releaseSpecificationPath}
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using a literal version, but their dependents via "peerDependencies" have no changes since their latest release and have their version specified as null in the release spec', async () => {
+    it('throws if there are any packages in the release with a major version bump using a literal version, but their dependents via "peerDependencies" have their version specified as null in the release spec, even if they have no changes', async () => {
       await withSandbox(async (sandbox) => {
         const project = buildMockProject({
           workspacePackages: {
--- a/src/release-specification.test.ts
+++ b/src/release-specification.test.ts
@@ -602,7 +602,7 @@ Your release spec could not be processed due to the following issues:
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "peerDependencies" are not listed in the release', async () => {
+    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "peerDependencies" have changes since their latest release and are not listed in the release', async () => {
       await withSandbox(async (sandbox) => {
         const project = buildMockProject({
           workspacePackages: {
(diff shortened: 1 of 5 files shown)


src/release-specification.test.ts
```diff
@@ -602,7 +602,7 @@ Your release spec could not be processed due to the following issues:
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "peerDependencies" are not listed in the release', async () => {
+    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "peerDependencies" have changes since their latest release and are not listed in the release', async () => {
```
mcmire, line 605: Ah, good change, this is definitely more clear.

src/release-specification.test.ts
```diff
@@ -657,7 +657,62 @@ ${releaseSpecificationPath}
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using a literal version, but any of their dependents defined as "peerDependencies" are not listed in the release', async () => {
+    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "peerDependencies" have no changes since their latest release and are not listed in the release', async () => {
```
mcmire, line 660: Nit: I wonder if we should reword this slightly to emphasize that usually a package that has no changes doesn't need to be included in the release, but in this case that doesn't matter:

``` suggestion
    it('throws if there are any packages in the release with a major version bump using the word "major", but any of their dependents defined as "peerDependencies" are not listed in the release, even if they have no changes', async () => {

File: src/release-specification.test.ts [hunk 3]
```diff
@@ -712,7 +767,62 @@ ${releaseSpecificationPath}
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using the word "major", but their dependents via "peerDependencies" have their version specified as null in the release spec', async () => {
+    it('throws if there are any packages in the release with a major version bump using a literal version, but any of their dependents defined as "peerDependencies" have no changes since their latest release and are not listed in the release', async () => {
```
mcmire, line 770: Similar as above:

``` suggestion
    it('throws if there are any packages in the release with a major version bump using a literal version, but any of their dependents defined as "peerDependencies" are not listed in the release, even if they have no changes', async () => {

File: src/release-specification.test.ts [hunk 4]
```diff
@@ -768,7 +878,63 @@ ${releaseSpecificationPath}
       });
     });
 
-    it('throws if there are any packages in the release with a major version bump using a literal version, but their dependents via "peerDependencies" have their version specified as null in the release spec', async () => {
+    it('throws if there are any packages in the release with a major version bump using the word "major", but their dependents via "peerDependencies" have no changes since their latest release and have their version specified as null in the release spec', async () => {
```
mcmire, line 881: Again, similar:

``` suggestion
    it('throws if there are any packages in the release with a major version bump using the word "major", but their dependents via "peerDependencies" have their version specified as null in the release spec, even if they have no changes', async () => {

File: src/release-specification.test.ts [hunk 5]
```diff
@@ -818,6 +984,62 @@ Your release spec could not be processed due to the following issues:
 
 The release spec file has been retained for you to edit again and make the necessary fixes. Once you've done this, re-run this tool.
 
+${releaseSpecificationPath}
+`.trim(),
+        );
+      });
+    });
+
+    it('throws if there are any packages in the release with a major version bump using a literal version, but their dependents via "peerDependencies" have no changes since their latest release and have their version specified as null in the release spec', async () => {
```
mcmire, line 993:
``` suggestion
    it('throws if there are any packages in the release with a major version bump using a literal version, but their dependents via "peerDependencies" have their version specified as null in the release spec, even if they have no changes', async () => {

File: src/ui/App.tsx [hunk 6]
```diff
@@ -67,16 +73,35 @@ function App() {
...
         if (!res.ok) {
           throw new Error(`Received ${res.status}`);
         }
         return res.json();
       })
       .then((data: Package[]) => {
+        const newPackageNames = new Set(data.map((pkg) => pkg.name));
+
+        // Only clean up selections if the package list actually changed
+        if (!setsAreEqual(previousPackages.current, newPackageNames)) {
```
mcmire, line 94: Does `isDeepStrictEqual` (from `util`) work here?

src/ui/App.tsx
```diff
@@ -67,16 +73,35 @@ function App() {
...
+          setSelections((prev) =>
```
mcmire, line 95: Ah, I think this name is confusing me a bit. At first I thought this referred to the packages that users had selected via the checkbox. But this refers to packages that are shown in the list, is that true?

It's probably fine now, but perhaps we can consider changing this name in a future PR.

mcmire (dismissed):
Little nits but none of them are blocking. This looks good!

src/ui/App.tsx
cryptodev-2s: It should but I wasn't able to use due to below issue

```
vite v6.2.0 building for production...
[plugin vite:resolve] Module "util" has been externalized for browser compatibility, imported by "/Volumes/Projects/consensys/create-release-branch/src/ui/App.tsx". See https://vite.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
✓ 241 modules transformed.
✗ Build failed in 413ms
error during build:
src/ui/App.tsx (5:9): "isDeepStrictEqual" is not exported by "__vite-browser-external", imported by "src/ui/App.tsx".
file: /Volumes/Projects/create-release-branch/src/ui/App.tsx:5:9

3: import { createRoot } from 'react-dom/client';
4: import { SemVer } from 'semver';
5: import { isDeepStrictEqual } from 'util';
            ^
6: import { ErrorMessage } from './ErrorMessage.js';
7: import { PackageItem } from './PackageItem.js';

Comment (mcmire) [reply src/ui/App.tsx:94]: Oh interesting. Alright, that makes sense — `util` is a Node-builtin, so maybe Vite doesn't support it by default.

File: src/ui/App.tsx (hunk 7 above)
Comment (cryptodev-2s) [reply src/ui/App.tsx]: I did change it to `releaseSelections`

Comment (mcmire) [reply src/ui/App.tsx]: Hmm, I think we should do this in a new PR since it touches a bunch of new places. There are already a couple of distinct changes in this PR and maybe we shouldn't try to add more things.

Comment (cryptodev-2s) [reply src/ui/App.tsx:95]: Yes make sense, I did rollback the changes

File: src/ui/App.tsx [hunk 8]
```diff
@@ -251,11 +276,11 @@ function App() {
...
 
   const handleBulkAction = (action: ReleaseType) => {
-    const newSelections = { ...selections };
+    const newReleaseSelections = { ...selections };
     selectedPackages.forEach((packageName) => {
-      newSelections[packageName] = action;
+      newReleaseSelections[packageName] = action;
     });
-    setSelections(newSelections);
+    setSelections(newReleaseSelections);
```
mcmire, line 283: Are these left over from the rename commit?

``` suggestion
    const newSelections = { ...selections };
    selectedPackages.forEach((packageName) => {
      newSelections[packageName] = action;
    });
    setSelections(newSelections);
```

cryptodev-2s: Yes indeed I missed this one

PR #203, opened 2026-06-01
feat: add --skip-changelog-update flag

cryptodev-2s:
## Summary

Adds an opt-in `--skip-changelog-update` CLI flag. When set, the tool:
- Skips auto-populating each package's `## Unreleased` → `### Uncategorized` section from git commits since the last release.
- Skips the accompanying `Initialize Release X.Y.Z` commit.

The use case: repos that maintain their changelog on the go (each PR adds its own categorized entry under `## Unreleased`) don't need the auto-population step, which otherwise appends a noisy `### Uncategorized` block on top of those hand-written entries and has to be cleaned up by hand.

Default behavior is unchanged. Repos that don't pass the flag get exactly what they get today.

The flag gates the same `if (firstRun)` block in both `followMonorepoWorkflow` (non-interactive) and `startUI` (interactive web UI), so behavior is consistent regardless of which mode the user invokes. It composes naturally with `--backport`.

## Test plan

- [ ] CI: `yarn build` succeeds
- [ ] CI: `yarn jest` — 275 tests pass
- [ ] CI: `yarn lint` — exits 0
- [ ] Functional test `'does not create an Initialize Release commit when --skip-changelog-update is passed'` in `src/functional.test.ts` verifies the end-to-end behavior by invoking the real CLI binary via `tsx`
- [ ] Unit test in `src/monorepo-workflow-operations.test.ts` directly asserts that `updateChangelogsForChangedPackages` is not called and that no `Initialize Release` commit is made, while the `Update Release` commit still happens
- [ ] Manual smoke test against a downstream repo that maintains its changelog on the go (optional)


cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

cryptodev-2s:
@metamaskbot publish-preview

